PR #3, opened 2015-04-08
new: GitLabApi.create(url, user, pass)

patrikbeno:


patrikbeno:
force updated: javadocs added; exception thrown as is (not wrapped)

PR #13, opened 2017-02-10
Add queryParam for getProjects()

jlafourc:
Hello,

We should do the same as getAllProjects() to retrieve all projects at once.

Julien

### DIFF
--- a/src/main/java/com/messners/gitlab/api/ProjectApi.java
+++ b/src/main/java/com/messners/gitlab/api/ProjectApi.java
@@ -31,7 +31,7 @@ public class ProjectApi extends AbstractApi {
      * @throws GitLabApiException
      */
     public List<Project> getProjects() throws GitLabApiException {
-        Response response = get(Response.Status.OK, null, "projects");
+        Response response = get(Response.Status.OK, UriComponent.decodeQuery("per_page=9999", true), "projects");
         return (response.readEntity(new GenericType<List<Project>>() {
         }));
     }


gmessner:
Thanks again for your contribution, however I will not be merging this change as in the next week or so will be adding the getStarredProjects() method and the following parameters along with a per page parameter to all list project methods:

archived - Limit by archived status
visibility - Limit by visibility public, internal, or private
order_by - Return projects ordered by id, name, path, created_at, updated_at, or last_activity_at fields. Default is created_at
sort - Return projects sorted in asc or desc order. Default is desc
search - Return list of authorized projects matching the search criteria
simple - Return only the ID, URL, name, and path of each project

gmessner:
Julien,

Just wanted to let you know that changes were recently made and merged that included equivalent changes to this merge request.

I also wanted to get your input on the following:

I have been working on an Iterator implementation that encapsulates the paging of all calls to the API that support paging.  For projects it would work like this:

```java
// Get a Pager instance that will page through the projects with 10 projects per page
Pager<Project> projectPager = gitlabApi.getProjectsApi().getProjectsPager(10);

// Iterate through the pages and print out the name and description
while (projectsPager.hasNext())) {

    List<Project> projects = projectsPager.next();
    for (Project project : projects) {
        System.out.println(project.getName() + " -: " + project.getDescription());
    }
}
```
Pager is a regular Iterator without remove() implemented.  It pages over the available lists of objects by issuing the appropriate call to the GitLab API.  It also has the following public non-Iterator methods:

```java
List<T> first();
int getItemsPerPage();
int getNumPages();
List<T> getPage(int pageNumber);
int getTotalItems();
List<T> last();
List<T> previous();
```
Would love to get your feedback on this proposal.

gmessner:
@jlafourc 
GitLab4J-API 4.3.0 has been released. This includes both the new Pager support and page/perPage support for all API calls that return a List.

jlafourc:
Hi, 

I'm so sorry! Busy times... I'll try to have a look maybe this week. Many thanks for this! 

Julien

PR #14, opened 2017-02-10
Add new method on RepositoryApi to get tree recursively

jlafourc:
Hello,

I added some code to get a tree recursively

Julien


gmessner:
Nice addition and the code looks good.

Thanks for the contribution.

PR #15, opened 2017-02-15
Added new method to repository API to get the archive of the repository

jlafourc:
Hello,

I added a new API to get the repository archive. It returns an InputStream so you can do whatever you want with it.

Julien


gmessner:
Thanks again for your contributions.  This is a useful addition.

PR #21, opened 2017-05-18
Added getIssues with page and per_page parameters

matteofacchinetti:
getIssues() unless parameter returns only first page of issues.

Override this method with page and per_page parameters,
provides a way to read all issues page by page.

Signed-off-by: Matteo Facchinetti <matteo.facchinetti@sirius-es.it>

### DIFF
--- a/src/main/java/org/gitlab4j/api/ProjectApi.java
+++ b/src/main/java/org/gitlab4j/api/ProjectApi.java
@@ -682,7 +682,7 @@ public class ProjectApi extends AbstractApi {
     }
 
     /**
-     * Get a list of project's issues.
+     * Get a list of project's issues. Only returns the first page
      *
      * GET /projects/:id/issues
      *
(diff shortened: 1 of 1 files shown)


gmessner:
I have been working on an Iterator implementation that encapsulates the paging of calls to the API that support paging.  For issues it would work like this:

```java
// Get a Pager instance that will page through the issues with 10 issues per page
Pager<Issue> issuesPager = gitlabApi.getProjectsApi().getIssuesPager(projectId, 10);

// Iterate through the pages and print out the ID and title
while (issuesPager.hasNext())) {

    List<Issue> issues = issuesPager.next();
    for (Issue issue : issues) {
        System.out.println("#" + issues.getId() + " - " + issue.getTitle());
    }
}
```
Pager is a regular Iterator without remove() implemented.  It pages over the available lists of objects by issuing the appropriate call to the GitLab API.  It also has the following public non-Iterator methods:

```java

List<T> first();
int getNumPages();
List<T> getPage(int pageNumber);
List<T> last();
List<T> previous();
```
Would this work for your use?

gmessner:
I went ahead a merged these changes.  Will be releasing a new version soon that will include the Pager I discussed above.

matteofacchinetti:
Great news.
This iterator is exactly what you need to handle the multi-page commands.
Personally I agree with your implementation and I can use it to read entire issues divided by page.

Regards,
Matteo

gmessner:
@matteofacchinetti 
GitLab4J-API 4.3.0 has been released. This includes both the new Pager support and page/perPage support for all API calls that return a List.

Issue #26, opened 2017-06-13
Group getProjects() allways return null


spingaroliver:
Group getProjects() function allways return null instead of group project list.

gmessner:
This has been fixed and verified.

Issue #25, opened 2017-06-13
Group getOwnerId() allways return null


spingaroliver:
Group getOwnerId() function allways return null instead of parent group id.

gmessner:
This has been fixed and verified.

PR #28, opened 2017-06-13
Fix Group model and GroupApi

spingaroliver:
fix Group model,
complete missing elements,
get Group projects via the GroupApi
fix issue #26 
fix issue #25


gmessner:
Thank you for identifying this issue and fixing.  That being said, these fixes have been manually applied directly to gitlab4j and have also implemented additional methods in the GroupApi to bring it fully compliant with v4.

This pull request will be closed without merging.

PR #55, opened 2017-07-28
getGroup overloaded to support path argument

djayard:
The group-details service supports taking in a group's path.
https://docs.gitlab.com/ce/api/groups.html#details-of-a-group


gmessner:

PR #81, opened 2017-09-23
add GitLab's (very limited) time tracking API

schakko:
At the moment, GitLab does not provide further time tracking abilities through the API. This PR adds the current time tracking API.
pom.xml version has been bumped to 4.5.2

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<groupId>org.gitlab4j</groupId>
 	<artifactId>gitlab4j-api</artifactId>
 	<packaging>jar</packaging>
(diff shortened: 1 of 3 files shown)


gmessner:
@schakko 
First off thank you for the contribution, very much appreciated.  I will be merging it as is.

Just one ask, in the future, do not bump the version in pom.xml, this will cause it to skip version 4.5.1 altogether.  The version automatically gets bumped in the release process.

gmessner:
@schakko
I have expanded upon your work and overloaded the methods to take a newly defined Duration class and seconds.  These updates are available in 4.6.0

PR #83, opened 2017-10-09
Add getTrace to JobApi

fgro93:
Hope it's okay that I just return a String here.


gmessner:
Thank you for the contribution.

fgro93:
The traces I analyze are around 125000 Characters long and have around 1500 line. Don't know if they have a max size, but the size depends on the output of each job.

gmessner:
@fgro93 
Thanks for the info. 
Because of the potential size, I'll add methods to get the trace as a file and as an InputStream in the next release.

PR #85, opened 2017-10-25
Get project users

LilAlex:
Get a list of all users with access to specific project. The result is different from get members request.
https://docs.gitlab.com/ce/api/projects.html#get-project-users


gmessner:
Related to the null pointer exception when user was not admin, where was this exception being thrown?

LilAlex:
1. Initialize GitLabApi object with non admin **access** token.
2. Get current user from user's api.
3. CurrentUser.isAdmin will be null instead of false.

gmessner:
Being null is the expected behavior when the caller is a non-admin user.  The GitLab REST endpoint does not return a value for is_admin if the caller is not an admion, see https://docs.gitlab.com/ee/api/users.html#for-user,   isAdmin being null allows you to know it was unset.   When you want to check the value do it as follows to avoid the NPE:

if (currentUser.getIsAdmin() == Boolean.TRUE)  {
}

Setting it to false could potentially cause problems, so you'll need to revert that change,

src/main/java/org/gitlab4j/api/ProjectApi.java
```diff
@@ -23,22 +23,15 @@
 
 package org.gitlab4j.api;
 
-import java.io.UnsupportedEncodingException;
-import java.net.URLEncoder;
-import java.util.List;
+import org.gitlab4j.api.GitLabApi.ApiVersion;
+import org.gitlab4j.api.models.*;
```
gmessner: Please do not use wildcard imports.

gmessner:
@LilAlex 
I see that you closed this MR, did you not want the get project users functionality added?

LilAlex:
@gmessner I'm experiencing some issues with git after updating to High Sierra. Wanted to resolve them first and then open another pull request.
Reinstalled command tools and it looks that everything works once again.
Thanks.

PR #89, opened 2017-10-27
Get active/blocked users

bobkovalex:
Filter users based on states: blocked and active.
https://docs.gitlab.com/ce/api/users.html#for-normal-users


gmessner:
@bobkovalex 

PR #95, opened 2017-11-21
Update README.md

DoktorAerzt:
You mist the gitLabApi Reference for the Notes part i think and it need to be an IID.


gmessner:
@DoktorAerzt 
Thank you for pointing this out, I have corrected the example by calling the parameter IID in the comment and have also removed the "Integer" text from both of the parameters of the example call.  SInce I made the change directly, this PR is no longer needed.

PR #96, opened 2017-11-23
Add a way to get only one member of a group

Xephi:


gmessner:
@Xephi

PR #100, opened 2017-11-28
Create User API has some problems, and I modified it

DmuMenDuo:
The field external is repeated and will return  error` Bad Request`, which took me several days to find, so I open a pull request. and I add a field `skip_confirmation` [API](http://docs.gitlab.com/ce/api/users.html#user-creation),I need the field in my project.


DmuMenDuo:
oh,The length of the password is greater than 6, otherwise the error will be reported.

gmessner:
@DmuMenDuo 
Thank you for fixing this issue.

PR #102, opened 2017-11-30
Allow to setup client properties

Fuud:
`clientConfig.getProperties()` returns immutable view of all properties.
To add properties we need to pass them into `clientConfig.property(name, value)`


gmessner:
@Fuud 
Thank you for the contribution.

PR #108, opened 2017-12-06
"ref_name" was changed to "ref" in RepositoryAPI v4

khvalIvan:
https://gitlab.com/gitlab-org/gitlab-ce/blob/8-16-stable/doc/api/repositories.md - optional "ref_name" for V3
https://docs.gitlab.com/ce/api/repositories.html - optional "ref" for V4


gmessner:
@khvalIvan 
Nice fix and well documented.  Thanks for the contribution.

PR #125, opened 2018-01-09
Rework exception handling to attach original exception.

tiaanl:
When an errors occurs in the JSON parsing of the WebHookManager then the exceptions is caught, but completely new exceptions are throws, throwing away any useful information about the original cause.  By creating the `GitlabApiException` with the original exception allows you to use `ex.getCause()` to get to the original data without changing which exceptions are throws from the method.

### DIFF
--- a/src/main/java/org/gitlab4j/api/webhook/WebHookManager.java
+++ b/src/main/java/org/gitlab4j/api/webhook/WebHookManager.java
@@ -1,7 +1,6 @@
 
 package org.gitlab4j.api.webhook;
 
-import java.io.IOException;
 import java.io.InputStreamReader;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
@@ -14,9 +13,6 @@ import org.gitlab4j.api.GitLabApiException;
 import org.gitlab4j.api.HookManager;
 import org.gitlab4j.api.utils.HttpRequestUtils;
 import org.gitlab4j.api.utils.JacksonJson;
-
-import com.fasterxml.jackson.core.JsonParseException;
-import com.fasterxml.jackson.databind.JsonMappingException;
 
 /**
  * This class provides a handler for processing GitLab WebHook callouts.
--- a/src/main/java/org/gitlab4j/api/webhook/WebHookManager.java
+++ b/src/main/java/org/gitlab4j/api/webhook/WebHookManager.java
@@ -97,11 +97,8 @@ public class WebHookManager extends HookManager {
 
             fireEvent(event);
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/gitlab4j/api/webhook/WebHookManager.java
```diff
@@ -98,22 +97,13 @@ public void handleEvent(HttpServletRequest request) throws GitLabApiException {
...
-        } catch (IOException ioe) {
-            errorMessage = ioe.getMessage();
-            LOG.warning("Error reading JSON data, error=" + errorMessage);
+        } catch (IOException e) {
+            LOG.warning("Error parsing JSON data, error=" + e.getMessage());
+            throw new GitLabApiException(e);
         } catch (Exception e) {
-            errorMessage = e.getMessage();
-            LOG.warning("Unexpected error reading JSON data, error=" + errorMessage);
+            LOG.warning("Unexpected error reading JSON data, error=" + e.getMessage());
```
gmessner: Since you have already simplified it by removing other catch blocks, can you make the logged message more generic and only capture Exception and remove the catch (IOException e) block?  Something like this:

``` LOG.warning("Unexpected error processing JSON data, error=" + e.getMessage());```

Comment (tiaanl) [reply src/main/java/org/gitlab4j/api/webhook/WebHookManager.java]: There is still a distinction between `IOException` which is the base of both `JsonParseException` and `JsonMappingException` and just a plain `Exception`.  So going through logs and seeing the first message tells me there was likely a parsing error, but seeing the latter message means something really unexpected happened, which i find useful.

Comment (gmessner) [reply src/main/java/org/gitlab4j/api/webhook/WebHookManager.java]: @tiaanl 
SInce your change sets the cause, can you not get that info from the exception returned by using getCause() ?

Comment (gmessner) [reply src/main/java/org/gitlab4j/api/webhook/WebHookManager.java]: How about the following, simplifies, but still gives the exact reason for the exception?
```
} catch (Exception e) {
    LOG.warning("Error processing JSON data, exception=" + e.getClass().getSimpleName() + ", error=" + e.getMessage());
    throw new GitLabApiException(e);
}
```

tiaanl: Yeah that'll work perfectly.  Can see info in logs and expect error programmatically.  Added commit.

gmessner:
@tiaanl 
Thanks for the contribution and working with me on this.

gmessner:
@tiaanl 
This has been released in the latest release of GitLab4J-API (4.7.15).

PR #131, opened 2018-01-18
Support New "Job Hook" Hook Renaming

### LINKED ISSUE #130: Outdated Webhook does not support "Job Hook" rename
"Build Hook" was renamed to "Job Hook" in GitLab v9.3.0.

https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/11287

This project's WebHookManager is still using the old "Build Hook" when validating the incoming request, so you get a ```GitLabException "Unsupported X-Gitlab-Event, event Name=Job Hook"``` for all job events.

armsnyder:
closes #130 

Validated by sending a test event from GitLab and it works.

### DIFF
--- a/src/main/java/org/gitlab4j/api/webhook/BuildEvent.java
+++ b/src/main/java/org/gitlab4j/api/webhook/BuildEvent.java
@@ -10,8 +10,10 @@ import org.gitlab4j.api.models.User;
 @XmlAccessorType(XmlAccessType.FIELD)
 public class BuildEvent implements Event {
 
-    public static final String X_GITLAB_EVENT = "Build Hook";
-    public static final String X_GITLAB_EVENT_2 = "Job Hook";
+    /** X-Gitlab-Event header value pre GitLab v9.3.0 */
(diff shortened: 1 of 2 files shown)


gmessner:
@armsnyder 
First off thank you for your contribution and this certainly needs to be fixed.  The fix looks good other than the name choice for:
```
 public static final String X_GITLAB_EVENT = "Build Hook";
 public static final String X_GITLAB_EVENT_2 = "Job Hook";
```
I'd like these to be a bit more descriptive since there is no longer a one to one relationship, could you change these to the following?
```
 public static final String BUILD_HOOK_X_GITLAB_EVENT = "Build Hook";
 public static final String JOB_HOOK_X_GITLAB_EVENT = "Job Hook";
```

armsnyder:
Absolutely. I'll throw in a comment as well since even that change doesn't fully explain what's going on.

gmessner:
@armsnyder 
Thanks for making the changes.  I've merged the PR, will do a release sometime later this week, have a few other changes that need to go in before the release.

PR #135, opened 2018-01-31
Adds missing "MANUAL" job status

newcron:
I noticed that the MANUAL status is not present in the JobStatus Enum. A Job will have that status, when it's waiting for a manual trigger (`when: manual` in the .gitlab-ci.yml). 

### Current Situation
As it is currently, when trying to fetch all jobs for a pipeline, such as: 

```bash
 curl  --header "PRIVATE-TOKEN: $SECRET" "https://$HOST/api/v4/projects/$PROJECT/pipelines/$PIPELINE/jobs" 
```

or 

```java
gitLabApi.getJobApi().getJobsForPipeline(projectId, pipelineId)
```

jobs that are waiting to be triggered manually will have a `null` status. Adding this enum value fixes this issue.

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/JobStatus.java
+++ b/src/main/java/org/gitlab4j/api/models/JobStatus.java
@@ -33,7 +33,7 @@ import com.fasterxml.jackson.annotation.JsonValue;
  */
 public enum JobStatus {
 
-    RUNNING, PENDING, SUCCESS, FAILED, CANCELED, SKIPPED;
+    RUNNING, PENDING, SUCCESS, FAILED, CANCELED, SKIPPED, MANUAL;
 
     private static JacksonJsonEnumHelper<JobStatus> enumHelper = new JacksonJsonEnumHelper<>(JobStatus.class);
 


gmessner:
While reviewing this I noticed that there are two enums defined for this.

org,gitlab4j.api.Constants.JobScope
org,gitlab4j.api.models.JobStatus

JobScope has all the correct values.  What I would like to do is remove "org,gitlab4j.api.models.JobStatus" and rename "org,gitlab4j.api.Constants.JobScope" to "org,gitlab4j.api.Constants.JobStatus" (making sure to fix all references)

Could you do this?  It's just a bad pattern to have to things representing the same type.

gmessner:
@newcron 
While reviewing this I noticed that there are two enums defined for this.

org,gitlab4j.api.Constants.JobScope
org,gitlab4j.api.models.JobStatus

JobScope has all the correct values.  What I would like to do is remove "org,gitlab4j.api.models.JobStatus" and rename "org,gitlab4j.api.Constants.JobScope" to "org,gitlab4j.api.Constants.JobStatus" (making sure to fix all references)

Could you do this?  It's just a bad pattern to have to things representing the same type.

gmessner:
@newcron 
After further thought, it might cause some users grief I the change I suggested was made, so I have merged this as is.

Thanks for your contribution.

PR #137, opened 2018-02-05
Fixed update file API response code.

istsergey:
Response code for the update file API have to be 200('Ok')
Proof: 
line 130 - https://gitlab.com/gitlab-org/gitlab-ce/blob/master/lib/api/files.rb 
line 108 - https://gitlab.com/gitlab-org/gitlab-ce/blob/master/lib/api/v3/files.rb


gmessner:
Good catch and thanks for the contribution.  I have merged and will do a release with some non-related changes later this week.

PR #138, opened 2018-02-05
changed comparison of Integer references using the == to equals() 

istsergey:


gmessner:
Another good catch,  Thanks again for your contribution.

PR #139, opened 2018-02-06
Fixed Double-Checked Locking by declaring fields to be volatile

istsergey:
Double-Checked Locking without volatile field can work incorrectly
http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html


gmessner:
@istsergey 
While it may not harm marking these volatile,  it is not needed as all the sub-API objects hold no state (they are immutable).  To make them truly immutable, just need to mark the single field in AbstractApi final, as follows:

```java
private final  GitLabApi gitLabApi;
```

istsergey:
@gmessner you are right. Please reject my pull request.

gmessner:
Closed per submitter's request.

PR #152, opened 2018-03-07
Project archiving and unarchiving

augi:
As described [in the documentation](https://docs.gitlab.com/ce/api/projects.html#archive-a-project).


gmessner:
@augi 
Thanks for the contribution

PR #153, opened 2018-03-10
Project Branches API

lamdaV:
The Gitlab docs list other methods of gathering information or updating protected branches.

https://docs.gitlab.com/ce/api/protected_branches.html#list-protected-branches


gmessner:
@lamdaV 
Great contribution.  Though I'd like to get your input/consideration on two aspects:
1) GitLab4J tries mimic/mirror the GitLab API docs as close as possible, should this be its own sub-API called ```ProtectedBranchesApi``` ?
2) I think that the approach of using the Branch class for this might not be the best approach.  Users of GitLab4J often ask why fields are often null, and when fetching info on protected branches most of the Branch class fields will be null, and when fetching Branch info the protected branch fields will be null.  Suggest creating ```ProtectedBranch``` just for this use.

lamdaV:
@gmessner 
1. If the intent is to mimic/mirror it, I think it will be a good idea to create a sub-api called `ProtectedBranchesApi`. It will also avoid the naming collision with `unprotectBranch`.
2. I agree. I will create a separate `ProtectedBranch` model for this.

src/test/java/org/gitlab4j/api/TestUtils.java
```diff
@@ -33,8 +32,7 @@ public static String getReaderContentAsString(Reader reader) throws IOException
     static {
 
         testProperties = new Properties();
-        File path = new File(System.getProperty("user.home"), "test-gitlab4j.properties");
```
lamdaV: https://stackoverflow.com/questions/5529532/how-to-get-a-test-resource-file

To keep the resource under the test directory of the project. If this is undesired, I can remove it.

src/test/resources/test-gitlab4j-sample.properties
```diff
@@ -0,0 +1,4 @@
+TEST_NAMESPACE=some-namespace
+TEST_PROJECT_NAME=some-project-name
+TEST_HOST_URL=some-gitlab-url
+TEST_PRIVATE_TOKEN=some-private-token
```
lamdaV: Goes along with the above comment. Gives a new contributor an idea of how to setup for testing.

README.md
```diff
@@ -155,6 +155,7 @@ The API has been broken up into sub APIs classes to make it easier to learn and
 &nbsp;&nbsp;[SessionApi](#sessionapi)<br/>
 &nbsp;&nbsp;[SystemHooksApi](#systemhooksapi)<br/>
 &nbsp;&nbsp;[UserApi](#userapi)
+&nbsp;&nbsp;[ProtectedBranchesApi](#protectedbranchesapi)
```
gmessner: Try to keep this list and the examples in alphabetical order.

src/test/java/org/gitlab4j/api/TestUtils.java
gmessner: A few things here:

1. Not a fan of having it down in a src/test/resource directory, it leads to the potential of secrets being accidentally checked in.  I would prefer it in the root of the project, and name it:
```
example-test-gitlab4j.properties
```

When moved to the root the code in TestUtils would need to look something like:
```
try (InputStream input = new FileInputStream("test-gitlab4j.properties")) {

```

2. Include test-gitlab4j.properties in the .gitignore file so it is not accidentally checked in.

3.  There are about 10 additional properties that need to be set to fully test things, I'll get you those when I have some time tomorrow afternoon.

Last but not least.  I am working on a GitLab server docker image that will be used to test GitLab4J, this way the test-gitlab4j.properties file can actually be created and left as is.  I expect to have this finished and checked in within the next 2-3 weeks.

gmessner:
@lamdaV 
You've done a lot of work on this and it is very close to being complete, what do you think about me just merging it now,   I will then add some updates, create a PR and have you review.  Sound like a plan?

lamdaV:
@gmessner Sounds good to me!

gmessner:
@lamdaV 
I submitted the PR, you'll find it here:
[Feature/protected branch and example test properties (#156)](https://github.com/gmessner/gitlab4j-api/pull/156)

PR #156, opened 2018-03-12
Feature/protected branch and example test properties

gmessner:
@lamdaV 
1. Added bean test for ProtectedBranch in TestGitLabApiBeans and fixed issues exposed by the bean test (AccessLevel was a String, not an Integer). 
2. Added all known properties to the example-test-gitlab4j.properties file with comments and added logic to the test classes to skip tests that do not have the required property values set.

### DIFF
--- a/example-test-gitlab4j.properties
+++ b/example-test-gitlab4j.properties
@@ -1,33 +1,33 @@
-# To test GitLab4J-API you must have access to a gitlab server.
 #
-# Copy this file to test-gitlab4j.properties and update the property
-# values to match the gitlab server you are testing against.
+# To test GitLab4J-API, you must have access to a GitLab server.
 #
-# The TEST_PRIVATE_TOKEN and TEST_USERNAME should belong to an admin user
+# Copy this file to test-gitlab4j.properties and set the property
+# values to match the GitLab server you are testing against.
 #
 
-# The following values must be set or testing will fail
-TEST_NAMESPACE=some-namespace
-TEST_PROJECT_NAME=some-project-name
-TEST_HOST_URL=some-gitlab-url
-TEST_PRIVATE_TOKEN=admin-private-token
-TEST_ACCESS_TOKEN=some-access-token
-TEST_USERNAME=admin-user
+# REQUIRED: The following values must be set or the tests will fail.
+# The TEST_PRIVATE_TOKEN and TEST_USERNAME must belong to an admin user. 
+TEST_PRIVATE_TOKEN=
+TEST_ACCESS_TOKEN=
+TEST_USERNAME=
+TEST_NAMESPACE=
+TEST_PROJECT_NAME=
+TEST_HOST_URL=
 
-# To test the GroupApi, set these to an existing group and group-project, and group member
+# OPTIONAL: To test the GroupApi, set these to an existing group name and group-project name,
+# and the username of a group member
 TEST_GROUP=
 TEST_GROUP_PROJECT=
 TEST_GROUP_MEMBER_USERNAME=
 
-# To test oauth2Login, set these properties
+# OPTIONAL: To test oauth2Login, set these properties
 TEST_LOGIN_USERNAME=
 TEST_LOGIN_PASSWORD=
 
-# To test sudo capability provide a username to sudo as
+# OPTIONAL: To test sudo capability provide a username to sudo as
 TEST_SUDO_AS_USERNAME=
 
-# To test using GitLab4J-API with a proxy, set the following properties
+# OPTIONAL: To test using GitLab4J-API with a proxy, set the following properties
 TEST_PROXY_URI=
 TEST_PROXY_USERNAME=
 TEST_PROXY_PASSWORD=
-
--- a/example-test-gitlab4j.properties
+++ b/example-test-gitlab4j.properties
@@ -1,5 +1,33 @@
-# Rename this file to test-gitlab4j.properties with proper property values.
+# To test GitLab4J-API you must have access to a gitlab server.
+#
+# Copy this file to test-gitlab4j.properties and update the property
+# values to match the gitlab server you are testing against.
+#
+# The TEST_PRIVATE_TOKEN and TEST_USERNAME should belong to an admin user
+#
+
+# The following values must be set or testing will fail
 TEST_NAMESPACE=some-namespace
 TEST_PROJECT_NAME=some-project-name
 TEST_HOST_URL=some-gitlab-url
-TEST_PRIVATE_TOKEN=some-private-token
+TEST_PRIVATE_TOKEN=admin-private-token
+TEST_ACCESS_TOKEN=some-access-token
+TEST_USERNAME=admin-user
+
+# To test the GroupApi, set these to an existing group and group-project, and group member
+TEST_GROUP=
+TEST_GROUP_PROJECT=
+TEST_GROUP_MEMBER_USERNAME=
+
+# To test oauth2Login, set these properties
+TEST_LOGIN_USERNAME=
+TEST_LOGIN_PASSWORD=
+
+# To test sudo capability provide a username to sudo as
+TEST_SUDO_AS_USERNAME=
+
+# To test using GitLab4J-API with a proxy, set the following properties
+TEST_PROXY_URI=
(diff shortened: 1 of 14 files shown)


example-test-gitlab4j.properties
```diff
@@ -1,5 +1,33 @@
...
+TEST_LOGIN_USERNAME=
+TEST_LOGIN_PASSWORD=
+
+# To test sudo capability provide a username to sudo as
+TEST_SUDO_AS_USERNAME=
+
+# To test using GitLab4J-API with a proxy, set the following properties
+TEST_PROXY_URI=
+TEST_PROXY_USERNAME=
+TEST_PROXY_PASSWORD=
```
lamdaV: Should we follow the above convention of having some stub string such as `some-proxy-password`?

src/test/java/org/gitlab4j/api/TestGitLabApi.java
```diff
@@ -77,6 +77,7 @@ public void testGetVersion() throws GitLabApiException {
     @Test
     public void testProxyConnection() throws GitLabApiException {
         assumeTrue(TEST_PROXY_URI != null && TEST_PROXY_USERNAME != null && TEST_PROXY_PASSWORD != null);
+        assumeTrue(TEST_PROXY_URI.length() > 0 && TEST_PROXY_USERNAME.length() > 0 && TEST_PROXY_PASSWORD.length() > 0);
```
lamdaV: Suggest using `TEST_PROXY_URI.isEmpty()` for length checks.

src/test/java/org/gitlab4j/api/TestGroupApi.java
```diff
@@ -83,10 +83,12 @@ public static void setup() {
                 problems += "Problem fetching test group, error=" + gle.getMessage() + "\n";
             }
 
-            try {
-                testUser = gitLabApi.getUserApi().getUser(TEST_GROUP_MEMBER_USERNAME);
-            } catch (GitLabApiException gle) {
-                problems += "Problem fetching test user, error=" + gle.getMessage() + "\n";
+            if (TEST_GROUP_MEMBER_USERNAME != null && TEST_GROUP_MEMBER_USERNAME.length() > 0) {
```
lamdaV:
``` java
if (TEST_GROUP_MEMBER_USERNAME != null && !TEST_GROUP_MEMBER_USERNAME.isEmpty()) { ... }
```

src/test/java/org/gitlab4j/api/TestProtectedBranchesApi.java
```diff
@@ -157,6 +156,6 @@ public void testProtectBranch() throws GitLabApiException {
         List<ProtectedBranch> branches = gitLabApi.getProtectedBranchesApi().getProtectedBranches(project.getId());
         assertNotNull(branches);
         assertTrue(branches.stream()
-                .anyMatch((protectedBranch) -> protectedBranch.getName().equals(TEST_BRANCH_NAME)));
+                .anyMatch((protectedBranch) -> branch.getName().equals(TEST_BRANCH_NAME)));
```
lamdaV: Correct me if I am wrong, but I think this should remain unchanged. I wrote this test to protect an unprotected branch, get a list of protected branches, and verify that the unprotected branch is now part of this list. 

By changing the name, the assertion will always return `true` as `branch` is effectively treated as a `final` variable (line 155) and should always be `TEST_BRANCH_NAME` by the time (line 159) executes. If it does not, (line 155) should have failed with an exception thrown.

src/test/java/org/gitlab4j/api/TestUserApi.java
```diff
@@ -149,7 +153,7 @@ public void testGetOptionalUser() throws GitLabApiException {
     @Test
     public void testSudoAsUser() throws GitLabApiException {
 
-        assumeTrue(TEST_SUDO_AS_USERNAME != null);
+        assumeTrue(TEST_SUDO_AS_USERNAME != null && TEST_SUDO_AS_USERNAME.length() > 0);
```
lamdaV: `isEmpty`

src/test/java/org/gitlab4j/api/TestProtectedBranchesApi.java
gmessner: Not sure how that got changed, fixed and pushed.

example-test-gitlab4j.properties
gmessner: As you imply, to be consistent it should be one way or the other.  Providing a value could lead to some confusion, (are these valid values?), so I suggest we remove all values, something like this:
```
#
# To test GitLab4J-API, you must have access to a GitLab server.
#
# Copy this file to test-gitlab4j.properties and set the property
# values to match the GitLab server you are testing against.
#

# REQUIRED: The following values must be set or the tests will fail.
# The TEST_PRIVATE_TOKEN and TEST_USERNAME must belong to an admin user. 
TEST_PRIVATE_TOKEN=
TEST_ACCESS_TOKEN=
TEST_USERNAME=
TEST_NAMESPACE=
TEST_PROJECT_NAME=
TEST_HOST_URL=

# OPTIONAL: To test the GroupApi, set these to an existing group name and group-project name,
# and the username of a group member
TEST_GROUP=
TEST_GROUP_PROJECT=
TEST_GROUP_MEMBER_USERNAME=

# OPTIONAL: To test oauth2Login, set these properties
TEST_LOGIN_USERNAME=
TEST_LOGIN_PASSWORD=

# OPTIONAL: To test sudo capability provide a username to sudo as
TEST_SUDO_AS_USERNAME=

# OPTIONAL: To test using GitLab4J-API with a proxy, set the following properties
TEST_PROXY_URI=
TEST_PROXY_USERNAME=
TEST_PROXY_PASSWORD=

```

src/test/java/org/gitlab4j/api/TestGitLabApi.java
gmessner: ```TEST_PROXY_URI.length() > 0```
        vs
```!TEST_PROXY_URI.isEmpty()```


When checking for truth, I prefer to use a statement that results in true, not something that has to be negated to produce true.  It's just a personal preference and is used throughout GitLab4J, so to change it in these instances would require a sweep through all the code to be consistent.

src/test/java/org/gitlab4j/api/TestGroupApi.java
gmessner: See my response above

src/test/java/org/gitlab4j/api/TestUserApi.java
gmessner: See my response above.

src/test/java/org/gitlab4j/api/TestProtectedBranchesApi.java
gmessner: Made another change to this.  After the revert to protectedBranch the ```branch``` variable was not referenced, and I realized that the contents of the returned protected branch were never really validated, by validating the name, it took care of the unused reference.

example-test-gitlab4j.properties
lamdaV: I agree with what you said about validity. My comment was more so on picking one convention over another.

src/test/java/org/gitlab4j/api/TestGitLabApi.java
lamdaV: The reasoning sounds good to me.

lamdaV:
@gmessner 
LGTM

example-test-gitlab4j.properties
gmessner: I figured that, I just pushed this change.

gmessner:
Thanks for all your feedback on this, and the initial contribution. 

BTW, I'll get to the RunnerApi submission when I get home from my day job tonight.

src/test/java/org/gitlab4j/api/TestGitLabApi.java
gmessner: @lamdaV 
FYI, when checking for empty instead of not empty I should be using isEmpty(), so I have gone through all the tests and replaced str.size() == 0 with str.isEmpty(), just thought I would let you know.

PR #159, opened 2018-03-18
Health Check Api

lamdaV:
https://docs.gitlab.com/ce/user/admin_area/monitoring/health_check.html

This is a mostly admin endpoint.


src/main/java/org/gitlab4j/api/GitLabApiClient.java
```diff
@@ -46,6 +46,7 @@
 
     private ClientConfig clientConfig;
     private Client apiClient;
+    private String baseUrl;
```
lamdaV: Added a field to keep track of the original url.

src/main/java/org/gitlab4j/api/GitLabApiClient.java
```diff
@@ -273,29 +274,44 @@ void setSudoAsId(Integer sudoAsId) {
...
             if (pathArg != null) {
-                url.append("/");
-                url.append(pathArg.toString());
+                urlBuilder.append("/");
+                urlBuilder.append(pathArg.toString());
             }
         }
-
-        return (new URL(url.toString()));
+        return urlBuilder.toString();
```
lamdaV: factored out url building logic so that it can be applied with the baseUrl as well.

src/main/java/org/gitlab4j/api/HealthCheckApi.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Get Health Checks from the liveness endpoint.
+     *
+     * GET /-/liveness
+     *
+     * @param token Health Status token
+     * @return LivenessHealthCheck instance
+     * @throws GitLabApiException if any exception occurs
+     * @deprecated
```
lamdaV: Gitlab docs state these have been deprecated. I added them in any way so that it is at least backward compatible during the deprecation process.

src/main/java/org/gitlab4j/api/models/HealthCheckStatus.java
```diff
@@ -0,0 +1,17 @@
+package org.gitlab4j.api.models;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+public class HealthCheckStatus {
+    private String status;
```
lamdaV: Ideally, these should be enums but I don't know all the possible values other than "ok". If you have any ideas, I would be happy to update this.

src/main/java/org/gitlab4j/api/models/HealthCheckStatusLabel.java
```diff
@@ -0,0 +1,17 @@
+package org.gitlab4j.api.models;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+public class HealthCheckStatusLabel {
+    private String shard;
```
lamdaV: The only extra label I had was "shard". If you have an idea of other labels, I am willing to add them.

Ideally, this should be an enum as well but I have not encountered other values while testing.

src/main/java/org/gitlab4j/api/HealthCheckApi.java
gmessner: @lamdaV 
There is no good reason to add methods to deprecated endpoints.  A method should only be deprecated if it was already implemented and is being replaced.  Since the method is brand new in this library, no one has ever used it and no need to add it, do not wish to encourage people to use deprecated methods.

gmessner:
@lamdaV 
I have avoided adding functionality that is not part of the GitLab API, this health check API is part of the admin dashboard and requires an entirely different access token than the API does.  This could possibly lead to confusion when the same access token doesn't work with the other API calls.

I guess what I'm saying is that I'm not sure this belongs here, though you might be able to convince me otherwise.

lamdaV:
@gmessner 
Thanks for taking a look.

If I remove the deprecated call, there would be no methods that take a new token. That is, there should be no confusion between api token and monitoring token as the preferred method of accessing this gitlab endpoint is through ip_whitelisting. This is further added in the javadoc upon `404` errors if the user is confused. 

As for your question on whether this belongs here or not, I do not think I have a convincing argument. I will elaborate on my desired usage. I am working on a project that integrates heavily with gitlab and several other components. I would like to create a monitoring dashboard that monitors all components. Since I am already using a gitlab wrapper, I would rather use the wrapper's abstractions for gitlab to query their health endpoints. This will help keep all application communication through gitlab localize through the library. The alternative, and in my opinion, less elegant solution is to use barebone java to do so and parse the results. Other alternatives would include adding dependencies which I think is overkill for one kind of query.

If your main concern is possible confusion, then perhaps we can work with the `GitlabApi` class and have a method called `getAdminApi()` in which it constructs a class `GitLabAdminApi` which wraps things such as `HealthCheckApi` and other admin level apis. This way the intent is clear to the user that using this api requires elevated permissions. 

Here is what it would look:
```java
gitlabApi.getAdminApi().getHealthCheckApi().getLiveness();
```

Let me know what you think.

gmessner:
I think you are on the right track with making the AdminApi separate, however I think that we should not make the AdminApi/HealthCheckApi part of the GitLabApi as it will require one to properly instantiate a GitLabApi instance to get to the AdminApi HealthCheckApi instance.  I have an idea on how to make it separate, when I get home tonight I will provide more info on that or maybe even create it.

lamdaV:
@gmessner Can I get a status update on this?

gmessner:
I have decided to go with your changes to GitLabApi and GitLabApiClient, however after reviewing the health_controller.rb source I believe your implementation can be greatly simplified.  I have started a code review.

You'll find the source for the health_controller at: https://gitlab.com/gitlab-org/gitlab-ce/blob/master/app/controllers/health_controller.rb

src/main/java/org/gitlab4j/api/HealthCheckApi.java
```diff
@@ -0,0 +1,80 @@
...
+
+public class HealthCheckApi extends AbstractApi {
+    public HealthCheckApi(GitLabApi gitLabApi) {
+        super(gitLabApi);
+    }
+
+    /**
+     * Get Health Checks from the liveness endpoint.
+     * Requires ip_whitelist
+     *
```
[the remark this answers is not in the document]
lamdaV: Good idea. I put the url for all methods doc strings.

lamdaV:
@gmessner Thanks for the feedback. I think I made the changes you suggested accordingly. Let me know if I misunderstood something or if you have any other changes you would like to make.

README.md
```diff
@@ -191,6 +192,13 @@ List<Event> events = gitLabApi.getEventsApi().getAuthenticatedUserEvents(null, n
 List<Group> groups = gitLabApi.getGroupApi().getGroups();
 ```
 
+#### HealthCheckApi
+```java
+// Get the liveness endpoint health check results.
+// Assumes ip_whitelisted
+LivenessHealthCheck healthCheck = gitLabApi.getHealthCheckApi().getLiveness();
```
gmessner: LivenessHealthCheck should be HealthCheckInfo

gmessner:
@lamdaV 
Changes are spot on, just one more change to README.md and you got it.  Thanks for your contribution and my apologies for taking so long to get to it

gmessner:
@lamdaV 
I added unit tests, made a few minor changes, and released 4.8.12. Thanks again for your contribution.

PR #167, opened 2018-04-03
Add a variation of the getCommits method that supports Paging and a path

swilson11:
Simple variation to make it easy to find all commits for a given file.

### DIFF
--- a/src/main/java/org/gitlab4j/api/CommitsApi.java
+++ b/src/main/java/org/gitlab4j/api/CommitsApi.java
@@ -181,6 +181,29 @@ public class CommitsApi extends AbstractApi {
     }
 
     /**
+     * Get a Pager of repository commits in a project
+     *
+     * GET /projects/:id/repository/commits
+     *
+     * @param projectId the project ID to get the list of commits for
+     * @param ref the name of a repository branch or tag or if not given the default branch
+     * @param since only commits after or on this date will be returned
+     * @param until only commits before or on this date will be returned
+     * @param itemsPerPage the number of Commit instances that will be fetched per page
+     * @param path the path to file of a project
+     * @return a Pager containing the commits for the specified project ID
+     * @throws GitLabApiException GitLabApiException if any exception occurs during execution
+     */
+    public Pager<Commit> getCommits(int projectId, String ref, Date since, Date until, int itemsPerPage, String path) throws GitLabApiException {
+        Form formData = new GitLabApiForm()
+                .withParam("ref_name", ref)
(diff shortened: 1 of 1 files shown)


src/main/java/org/gitlab4j/api/CommitsApi.java
```diff
@@ -180,6 +180,29 @@ public CommitsApi(GitLabApi gitLabApi) {
         return (new Pager<Commit>(this, Commit.class, itemsPerPage, formData.asMap(),  "projects", projectId, "repository", "commits"));
```
swilson11: I could also modify this method to call the new one that I added, with null for the path, if you would prefer.

gmessner:
@swilson11 
Thank you for the contribution. I've added an inline comment.  Additionally, could you add the following methods?  It will make the additions consistent with the rest of the API:
```
public List<Commit> getCommits(int projectId, String ref, 
        Date since, Date until, String path) throws GitLabApiException

public List<Commit> getCommits(int projectId, String ref,
        Date since, Date until, String path, int page, int perPage) throws GitLabApiException
```

src/main/java/org/gitlab4j/api/CommitsApi.java
swilson11: Done.

src/main/java/org/gitlab4j/api/CommitsApi.java
```diff
@@ -180,6 +180,29 @@ public CommitsApi(GitLabApi gitLabApi) {
...
+     * @param projectId the project ID to get the list of commits for
+     * @param ref the name of a repository branch or tag or if not given the default branch
+     * @param since only commits after or on this date will be returned
+     * @param until only commits before or on this date will be returned
+     * @param itemsPerPage the number of Commit instances that will be fetched per page
+     * @param path the path to file of a project
+     * @return a Pager containing the commits for the specified project ID
+     * @throws GitLabApiException GitLabApiException if any exception occurs during execution
+     */
+    public Pager<Commit> getCommits(int projectId, String ref, Date since, Date until, int itemsPerPage, String path) throws GitLabApiException {
```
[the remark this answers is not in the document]
swilson11: Done

swilson11:
For some reason, I can't reply to your comment above about adding the 2 new APIs. I added the second one, but the first one was already there. Thanks for maintaining this project, and for your quick responses on my request!

Sue

gmessner:
Changes look good, and thanks again for your contribution.

gmessner:
@swilson11 
This has been released in 4.8.9

PR #172, opened 2018-04-10
 Implemented being able to set a tag_list when creating and updating a project.

jimrinhealthcare:
I've left a note in for the V3 API section as I wasn't sure if you'd prefer to just ignore tag lists here (Gitlab doesn't appear to do anything with them if you send them) or throw some sort of exception.

I've added what I think the unit test changes should be, but couldn't work out what needed setting up to run them.


gmessner:
@jimrinhealthcare 
Thank you for your contribution.  I am going to go ahead and merge this as 4.8.10 is being ready for release and it would be good to have this in the release.  I will add the exception after it is merged.

PR #173, opened 2018-04-13
using post method instead of get to enable a runner

fouadh:
Hi,

This fix changes the enableRunner() method of RunnersApi to use the http POST method instead of GET to perform the operation.

Best regards,
Fouad


gmessner:
@fouadh 
Thank you for your contribution.  This will be available in the next release which will happen in the next couple of days.

PR #183, opened 2018-05-10
Add personal snippets api

faustovaz:
This adds support for Gitlab Personal Snippets API (/snippets). If everything is ok, it would be nice to have this code merged to the main project.

### DIFF
--- a/src/main/java/org/gitlab4j/api/GitLabApi.java
+++ b/src/main/java/org/gitlab4j/api/GitLabApi.java
@@ -71,7 +71,7 @@ public class GitLabApi {
     private LabelsApi labelsApi;
     private NotesApi notesApi;
     private EventsApi eventsApi;
-    private PersonalSnippetApi personalSnippetApi;
+    private SnippetApi snippetApi;
 
     /**
      * Get the GitLab4J shared Logger instance.
@@ -1279,20 +1279,20 @@ public class GitLabApi {
     }
     
     /**
-     * Gets the PersonalSnippetApi instance owned by this GitLabApi instance. The PersonalSnippetApi is used
-     * to perform all personal snippet related API calls.
+     * Gets the SnippetApi instance owned by this GitLabApi instance. The SnippetApi is used
+     * to perform all snippet related API calls.
      *
-     * @return the PersonalSnippetApi instance owned by this GitLabApi instance
+     * @return the SnippetApi instance owned by this GitLabApi instance
      */
-	public PersonalSnippetApi getPersonalSnippetApi() {
-        if (personalSnippetApi == null) {
+	public SnippetApi getSnippetApi() {
+        if (snippetApi == null) {
             synchronized (this) {
-                if (personalSnippetApi == null) {
-                	personalSnippetApi = new PersonalSnippetApi(this);
+                if (snippetApi == null) {
+                	snippetApi = new SnippetApi(this);
                 }
             }
         }
 
-        return personalSnippetApi;
+        return snippetApi;
 	}
 }
--- a/src/main/java/org/gitlab4j/api/GitLabApi.java
+++ b/src/main/java/org/gitlab4j/api/GitLabApi.java
@@ -71,7 +71,7 @@ public class GitLabApi {
     private LabelsApi labelsApi;
     private NotesApi notesApi;
     private EventsApi eventsApi;
-    private SnippetsApi snippetApi;
+    private SnippetsApi snippetsApi;
 
     /**
      * Get the GitLab4J shared Logger instance.
@@ -1279,20 +1279,20 @@ public class GitLabApi {
     }
     
     /**
-     * Gets the SnippetApi instance owned by this GitLabApi instance. The SnippetApi is used
+     * Gets the SnippetsApi instance owned by this GitLabApi instance. The SnippetsApi is used
      * to perform all snippet related API calls.
      *
-     * @return the SnippetApi instance owned by this GitLabApi instance
+     * @return the SnippetsApi instance owned by this GitLabApi instance
      */
 	public SnippetsApi getSnippetApi() {
-        if (snippetApi == null) {
+        if (snippetsApi == null) {
             synchronized (this) {
-                if (snippetApi == null) {
-                	snippetApi = new SnippetsApi(this);
+                if (snippetsApi == null) {
+                	snippetsApi = new SnippetsApi(this);
                 }
             }
         }
 
-        return snippetApi;
+        return snippetsApi;
 	}
 }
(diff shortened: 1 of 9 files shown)


gmessner:
@faustovaz 
This looks great and is very complete, thank you.

I have just one question, why call the the ```PersonalSnippetsApi``` and not simply the ```SnippetsApi```?  As much as possible, we keep a one-to-one mapping of sub-API names to the API names in the GitLab documentation, and this maps to https://docs.gitlab.com/ee/api/snippets.html which GitLab refers to as the **Snippets API**.

I would suggest renaming this to simply ```SnippetsApi``` so that it matches the GitLab documentation.

faustovaz:
Hi @gmessner ,
you are right. I chose PersonalSnippets to differentiate this kind of Snippet from Project's Snippet. But it's better to keep a one-to-one mapping, like you said.
I already submitted the code changes. 
Thank you!

src/main/java/org/gitlab4j/api/SnippetsApi.java
```diff
@@ -0,0 +1,125 @@
...
+     * @param fileName the file name of the snippet
+     * @param content the content of the snippet
+     * @return the created Snippet
+     * @throws GitLabApiException if any exception occurs
+     */
+	public Snippet createSnippet(String title, String fileName, String content) throws GitLabApiException {
+		GitLabApiForm formData = new GitLabApiForm()
+				.withParam("title", title)
+				.withParam("file_name", fileName)
+				.withParam("content", content);
```
gmessner: The setting ```title```, ```file_name```, and ```content``` must have a boolean third paramter specifying that they are required per the docs at: https://docs.gitlab.com/ee/api/snippets.html#create-new-snippet

Should look like this:
```
GitLabApiForm formData = new GitLabApiForm()
    .withParam("title", title, true)
    .withParam("file_name", fileName, true)
    .withParam("content", content, true);
```

src/main/java/org/gitlab4j/api/GitLabApi.java
```diff
@@ -71,6 +71,7 @@ public String getApiNamespace() {
     private LabelsApi labelsApi;
     private NotesApi notesApi;
     private EventsApi eventsApi;
+    private SnippetsApi snippetApi;
```
gmessner: Rename ```snippetApi``` to ```snippetsApi```

src/main/java/org/gitlab4j/api/SnippetsApi.java
```diff
@@ -0,0 +1,125 @@
...
+     * @param visibility the visibility (Public, Internal, Private) of the snippet
+     * @param description the description of the snippet
+     * @return the created Snippet
+     * @throws GitLabApiException if any exception occurs
+     */
+	public Snippet createSnippet(String title, String fileName, String content, Visibility visibility, String description) throws GitLabApiException {
+		GitLabApiForm formData = new GitLabApiForm()
+				.withParam("title", title)
+				.withParam("file_name", fileName)
+				.withParam("content", content)
```
gmessner: See above about ```title```, ```file_name```, and ```content``` being required parameters.

src/main/java/org/gitlab4j/api/SnippetsApi.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * Get a list of Authenticated User's Snippets.
+     *
+     * GET /snippets
+     *
+     * @return a list of authenticated user's snippets
+     * @throws GitLabApiException if any exception occurs
+     */
+	public List<Snippet> getSnippets() throws GitLabApiException {
```
gmessner: Add a parameter to this method indicating where to download the content, i.e.:
```
public List<Snippet> getSnippets(boolean downloadContent) throws GitLabApiException {
```
and the make fetching the content optional based on this.  You could add another method without the parameter that calls this method with ```downloadContent = false``` . The default should be false to make it act like the GitLab Snippets API.

src/main/java/org/gitlab4j/api/SnippetsApi.java
```diff
@@ -0,0 +1,125 @@
...
+
+	/**
+	 * Get a specific Snippet
+	 * 
+	 * @param snippetId the snippet ID to remove 
+	 * @return the snippet with the given id
+	 * @throws GitLabApiException if any exception occurs
+	 */
+	public Snippet getSnippet(Integer snippetId) throws GitLabApiException {
+		if (snippetId == null) {
```
gmessner: Do the same here with a flag on whether to download the content or not.

src/test/java/org/gitlab4j/api/TestSnippetsApi.java
```diff
@@ -0,0 +1,132 @@
...
+	public void testList() throws GitLabApiException {
+		Snippet snippet1 = createSnippet(new Snippet("Snippet 1", "Snippet.java", "Java content"));
+		Snippet snippet2 = createSnippet(new Snippet("Snippet 2", "Snippet.java", "Another java content"));
+		
+		SnippetsApi api = gitLabApi.getSnippetApi();
+		List<Snippet> snippets = api.getSnippets();
+		
+		assertTrue(snippets.size() >= 2);
+		assertTrue(snippets.stream().anyMatch(s -> s.getContent().equals("Java content")));
+		assertTrue(snippets.stream().anyMatch(s -> s.getContent().equals("Another java content")));
```
gmessner: When comparing strings in unit tests it is a good idea to use constants, so you are certain it is the same when you do the assert below.

src/test/java/org/gitlab4j/api/TestSnippetsApi.java
```diff
@@ -0,0 +1,132 @@
...
+
+		
+		deleteSnippet(snippet1);
+		deleteSnippet(snippet2);
+	}
+	
+	@Test
+	public void testSnippetContent() throws GitLabApiException {
+		Snippet snippet = createSnippet(
+				new Snippet("Snippet 1", "Snippet.java", "System.out.println(\"\");"));
```
gmessner: Create a constant to hold the content and use that instead of literals:
```final String CONTENT = "System.out.println(\"\");";```

File: src/test/java/org/gitlab4j/api/TestSnippetsApi.java [hunk 8]
```diff
@@ -0,0 +1,132 @@
...
+	
+	@Test
+	public void testRetrieveSnippet() throws GitLabApiException {
+		Snippet snippet = createSnippet (new Snippet(
+													"Xml Snippet", 
+													"file.xml", 
+													"<parent><data>1</data></parent>", 
+													Visibility.INTERNAL, 
+													"Description"));
+
```
gmessner: Same recommendation for using a constant for the content.

faustovaz:
Thank you for reviewing my code, @gmessner .
Now it's much better!

gmessner:
@faustovaz 
The changes look good.  I'm going to approve and merge and will get a release out later today.

gmessner:
@faustovaz 
I've released 4.8.19. 

I did make some minor edits, mostly around formatting to bring the new files in-line with the formatting styles we use (mainly 4 spaces instead of tabs).  Also added Pager and Optional support to the SnippetsApi.

Issue #180, opened 2018-05-02
Repositories contributors cannot be retrieved


ISibboI:
The api seems to be missing a method to retrieve a projects contributors, using this call:

https://docs.gitlab.com/ee/api/repositories.html#contributors

Is there any reason for that? Otherwise I'd just create a pull request.

gmessner:
This is a relatively new endpoint.  If you'd like to create a PR that would be great.

If you are unable to submit a PR, I can get this added and released in the next release of GitLab4J which is scheduled for 5/7/2018.

ISibboI:
Awesome! I will probably add multiple features since I need this library for a company project, and it is a great foundation.

gmessner:
@ISibboI 
Are you still planning on creating a PR with your enhancements, or should I go ahead and implement this specific one?

gmessner:
@ISibboI 
Thanks for contributing the enhancement.

PR #186, opened 2018-05-24
Get more information from gitlab

ISibboI:
Hey,

as mentioned in issue #180 before, here is a pull request with additions to the API. I added more than just the one missing method in this. I hope it's fine to merge it alltogether.

Additions:
+ get commit comments
+ get merge request changes: basically a diff of the merge request
+ get merge request participants
+ get project forks
+ get repository contributors

Fixes:
+ getFile method encoded spaces in file paths as + and not %F0, as gitlab wants.

### DIFF
--- a/src/main/java/org/gitlab4j/api/AbstractApi.java
+++ b/src/main/java/org/gitlab4j/api/AbstractApi.java
@@ -3,7 +3,6 @@ package org.gitlab4j.api;
 import java.io.File;
 import java.net.URL;
 import java.net.URLEncoder;
-import java.nio.charset.StandardCharsets;
 
 import javax.ws.rs.NotAuthorizedException;
 import javax.ws.rs.core.Form;
--- a/src/main/java/org/gitlab4j/api/AbstractApi.java
+++ b/src/main/java/org/gitlab4j/api/AbstractApi.java
@@ -3,6 +3,7 @@ package org.gitlab4j.api;
 import java.io.File;
 import java.net.URL;
 import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 
 import javax.ws.rs.NotAuthorizedException;
 import javax.ws.rs.core.Form;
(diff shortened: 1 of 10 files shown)


src/main/java/org/gitlab4j/api/AbstractApi.java
```diff
@@ -88,6 +88,31 @@ protected String urlEncode(String s) throws GitLabApiException {
...
+     * 
+     * Standard URL encoding changes spaces to plus signs, but for arguments that are part of the path,
+     * like the :file_path in a "Get raw file" request, gitlab expects spaces to be encoded with %20.
+     * 
+     * @param s the string to encode
+     * @return encoded version of s with spaces encoded as %2F
+     * @throws GitLabApiException if encoding throws an exception
+     */
+    protected String urlEncodeForPath(String s) throws GitLabApiException {
+        try {
```
gmessner: Why not just add `encoded = encoded.replace("+", "%20");` to `urlEncode(String s)`

I'm pretty sure  `urlEncode(String s)` not processing `+` is an oversight

src/main/java/org/gitlab4j/api/MergeRequestApi.java
```diff
@@ -577,4 +578,34 @@ public MergeRequest unapproveMergeRequest(Integer projectId, Integer mergeReques
...
+     * 
+     * GET /projects/:id/merge_requests/:merge_request_iid/participants
+     * 
+     * @param projectId the project ID to get the merge requests for
+     * @param mergeRequestIid the IID of the merge request to get
+     * @param itemsPerPage the number of Participant instances that will be fetched per page
+     * @return all participants for the specified merge request
+     * @throws GitLabApiException if any exception occurs
+     */
+    public Pager<Participant> getParticipants(Integer projectId, Integer mergeRequestIid, int itemsPerPage) throws GitLabApiException {
```
gmessner: Should be returning `Pager<User>`  no need to create a new class for the results.

src/main/java/org/gitlab4j/api/RepositoryFileApi.java
```diff
@@ -43,7 +43,7 @@ public RepositoryFile getFile(String filePath, Integer projectId, String ref) th
         Form form = new Form();
         addFormParam(form, "ref", ref, true);
         Response response = get(Response.Status.OK, form.asMap(),
-                "projects", projectId, "repository", "files", urlEncode(filePath));
+                "projects", projectId, "repository", "files", urlEncodeForPath(filePath));
```
gmessner: Change back to `urlEncode(filePath)` after fix to AbstractApi.urlEncode(String s)

src/main/java/org/gitlab4j/api/models/Contributor.java
```diff
@@ -0,0 +1,7 @@
+package org.gitlab4j.api.models;
+
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement
+public class Contributor extends AbstractUser {
```
gmessner: No need for this class, use `User` instead

src/main/java/org/gitlab4j/api/models/Participant.java
```diff
@@ -0,0 +1,7 @@
+package org.gitlab4j.api.models;
+
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement
+public class Participant extends AbstractUser {
```
gmessner: No need for this class, use `User` instead

src/main/java/org/gitlab4j/api/RepositoryApi.java
```diff
@@ -498,4 +499,19 @@ public CompareResults compare(String projectPath, String from, String to) throws
...
+     *
+     * GET /projects/:id/repository/contributors
+     *
+     * @param projectId the project to get the list of contributors for
+     * @param itemsPerPage the number of Project instances that will be fetched per page
+     * @return the list of contributors for the specified project ID
+     *
+     * @throws GitLabApiException if any exception occurs
+     */
+    public Pager<Contributor> getContributors(Integer projectId, int itemsPerPage) throws GitLabApiException {
```
gmessner: Should be returning `Pager<User>`  no need to create a new class for the results.

gmessner:
After further thought and examination of the other classes that sub-classed AbstractUser, I have changed my mind and deleted the comments about not needing Contributor and Participant classes.  I know believe it is correct creating those classes.  Address the urlEncode() change and this is good to merge.

ISibboI:
Thank you for your feedback! I was not sure what the urlEncode method exactly does, so I first implemented a second one, to be safe. Now it should be as you requested.

gmessner:
@ISibboI 
This is now available n release 4.8.20

PR #191, opened 2018-06-02
Added support for Wikis API

### LINKED ISSUE #189: Wikis API
This library lacks support for [Wikis API](https://docs.gitlab.com/ee/api/wikis.html).
I may submit a PR to add this feature.

shuklaalok7:
It fixes #189. Please review/discuss and merge these changes.


shuklaalok7:
I have added tests as were written in `TestProjectApiSnippets` class. All the tests are passing.

src/main/java/org/gitlab4j/api/ProjectApi.java
```diff
@@ -1843,6 +1833,139 @@ public String getRawSnippetContent(Integer projectId, Integer snippetId) throws
         }
     }
 
+    /**
```
gmessner: The convention we use is to have a one-to-one mapping of an API class to the GitLab API documentation page, in this case https://docs.gitlab.com/ce/api/wikis.html.  The code that was added to the ```ProjectApi``` class should be moved to a new class called WikisApi, you'll then need to add a ```public WikisApi getWikisApi()``` method to the ```GitLabApi ```class.

shuklaalok7:
@gmessner Thanks for reviewing it.
Actually, I had added the `WikisApi` class, but later removed when I observed following pattern in the code.
1. There is a `SnippetsApi` class for `/snippets` endpoints.
2. Further implementation for the `/projects/:id/snippets` endpoints is added in the `ProjectApi` class.

As the endpoints for Wikis API are in-line with the second case, I added support for this API in `ProjectApi` class and removed `WikisApi` class. The endpoints for Wikis API are `/projects/:id/wikis`. If you still suggest that I bring back `WikisApi` class, I'll do that this week.

src/main/java/org/gitlab4j/api/ProjectApi.java
gmessner: The project snippets documentation use to be part of the Projects API at GitLab, it still is, but they moved the documentation to it's own page at https://docs.gitlab.com/ee/api/project_snippets.html

Notice that this is not called an API it is called "Project snippets". That being said, I will eventually deprecate the snippet methods in the ```ProjectApi``` and relocate them to ```SnippetsApi```.  You will also find a few others similar to this, I deprecate and relocate when I am making mods in the file for some other reason.

The GitLab API has evolved over the years, it is almost impossible to keep up with the changes.  But keeping a one-to-one mapping of API classes to API documentation pages is how this project does it now, so please use the ```WikisApi``` approach.

shuklaalok7:
@gmessner Requested changes have been made. Let me know if any more change is needed. Tests are passing well.

shuklaalok7:
@gmessner Thanks.

PR #192, opened 2018-06-03
add project statistics information to projects api

jgoldhammer:
Currently the projects api can be used to request project statistics from gitlab, but the response for the statistics is not recognized. This PR adds the information to the projects class.


jgoldhammer:
@gmessner any thoughts?

src/main/java/org/gitlab4j/api/models/ProjectStatistics.java
```diff
@@ -0,0 +1,73 @@
...
+    public void setLfsObjectSize(long lfsObjectSize) {
+        this.lfsObjectSize = lfsObjectSize;
+    }
+
+    public void setStorageSize(long storageSize) {
+        this.storageSize = storageSize;
+    }
+
+    @Override
+    public String toString() {
```
gmessner: Would prefer to not override toString(), as it would be inconsistent with the other model classes.  If one would like to get a String representation of the instance you can call the following, which will return formatted JSON:

```
String json = org.gitlab4j.api.utils.JacksonJson.marshal(objectInstance);
```

src/main/java/org/gitlab4j/api/models/Project.java
```diff
@@ -592,7 +593,69 @@ public Project withPrintingMergeRequestLinkEnabled(Boolean printingMergeRequestL
...
+
+    public void setStatistics(ProjectStatistics statistics) {
+        this.statistics = statistics;
+    }
+
     public static final boolean isValid(Project project) {
         return (project != null && project.getId() != null);
     }
+
+    @Override
```
gmessner: Would prefer to not override toString(), as it would be inconsistent with the other model classes.  If one would like to get a String representation of the instance you can call the following, which will return formatted JSON:

```
String json = org.gitlab4j.api.utils.JacksonJson.marshal(objectInstance);
```

jgoldhammer:
I have updated my PR. I hope, it is ok now. 

Thanks
Jens

PR #196, opened 2018-06-07
updateUser bug

zhengrenjie:
the name of skip confirmation param of update user should be skip_reconfirmation instead of skip_confirmation.


gmessner:
@zhengrenjie 
Thanks for the fix.  It has been merged and will be released later today.

PR #204, opened 2018-06-27
added a pom dependency: fastjson and added a method to get languages used in a project

zhengrenjie:


zhengrenjie:
Added a pom dependency: fastjson and added a method to get languages used in a project.
Added license template method and model.
Added markdown transformed method and model.

src/main/java/org/gitlab4j/api/ProjectApi.java
```diff
@@ -2176,4 +2177,23 @@ public Project unstarProject(Object projectIdOrPath) throws GitLabApiException {
         Response response = post(expectedStatus, (Form) null, "projects", getProjectIdOrPath(projectIdOrPath), "unstar");
         return (response.readEntity(Project.class));
     }
+
```
gmessner: No need to introduce a new type for this, please re-implement `getProjectLanguage(Integer projectId)` as follows:
```
    /**
     * Get languages used in a project with percentage value.
     *
     * Get /projects/:id/languages
     *
     * @param projectId the ID of the project
     * @return a Map instance with the language as the key and the percentage as the value
     * @since GitLab 10.8
     */
    public Map<String, Float> getProjectLanguages(Integer projectId) throws GitLabApiException {

        if (projectId == null) {
            throw new RuntimeException("projectId cannot be null");
        }

        Response response = get(Response.Status.OK, null, "projects", projectId, "languages");
        return (response.readEntity(new GenericType<Map<String, Float>>() { }));
    }
```

pom.xml
```diff
@@ -206,6 +206,11 @@
 			<artifactId>javax.activation</artifactId>
 			<version>1.2.0</version>
 		</dependency>
+    <dependency>
+      <groupId>com.alibaba</groupId>
+      <artifactId>fastjson</artifactId>
+      <version>1.2.47</version>
+    </dependency>
```
gmessner: No need to introduce this new library, see my comments in ProjectApi.java

zhengrenjie:
@gmessner Thank you very much for your suggestion. I will submit it again after the revision.

zhengrenjie:
I have completed the changes.if there are any questions please tell me, I'll fix them.

gmessner:
Great job and making the requested changes.

gmessner:
@zhengrenjie 
This PR has been released ibn release 4.8.27. Thanks again for your contribution.

PR #206, opened 2018-06-27
Adds a new updateMergeRequest method with more attributes

### LINKED ISSUE #205: Missing Attributes in Update Merge Request API
There are a number of attributes in the Update Merge Request API which cannot be set using the existing parameters of the MergeRequestApi#updateMergeRequest method.

The full list of accepted attributes can be found here:
https://docs.gitlab.com/ee/api/merge_requests.html#update-mr

armsnyder:
I deprecated the existing updateMergeRequest method and added a new updateMergeRequest method which accepts more parameters. These new parameters correspond to the attributes [listed here](https://docs.gitlab.com/ee/api/merge_requests.html#update-mr) that were previously not settable.

Closes #205 

Since there were now 3 different updateMergeRequest methods which contained duplicated code, I abstracted the common logic to a protected method

```java
protected MergeRequest updateMergeRequest(Integer projectId, Integer mergeRequestIid, Form formData)
```
The reason it is protected and not private is so that in the future, users of this library can extend the class to add new attributes themselves, instead of waiting for the library to update. I'm curious if making it a public method would be preferable.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,8 @@
 		<jackson.version>2.9.3</jackson.version>
 		<javaServlet.version>3.1.0</javaServlet.version>
 		<junit.version>4.12</junit.version>
+		<mockito.version>1.10.19</mockito.version>
+		<hamcrest.version>1.3</hamcrest.version>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 	</properties>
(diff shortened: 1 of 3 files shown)


gmessner:
This functionality was already added by way of the MergeRequestFilter class which provides a fluent API for all the possible parameters.  See the following methods in MergeRequestApi:
```
public List<MergeRequest> getMergeRequests(MergeRequestFilter filter)
public List<MergeRequest> getMergeRequests(MergeRequestFilter filter, int page, int perPage)
public Pager<MergeRequest> getMergeRequests(MergeRequestFilter filter, int itemsPerPage)
```

gmessner:
@armsnyder 
Thanks for the contribution, looks good and I always appreciate unut tests being included.

PR #209, opened 2018-07-03
Add download a single artifact file endpoint

mpkorstanje:
[Download a single artifact file ](https://docs.gitlab.com/ee/api/jobs.html#download-a-single-artifact-file)

Introduced in GitLab 10.0

Download a single artifact file from within the job's artifacts archive.
Only a single file is going to be extracted from the archive and streamed to a client.


gmessner:
@mpkorstanje 
Thank you for the contribution.  I will create a release and let you know when it is completed.

gmessner:
@mpkorstanje 
This has been released in 4.8.28

PR #211, opened 2018-07-03
Add Release Api

eutkin:
https://docs.gitlab.com/ee/api/tags.html#create-a-new-release
https://docs.gitlab.com/ee/api/tags.html#update-a-release

### DIFF
--- a/src/main/java/org/gitlab4j/api/CommitsApi.java
+++ b/src/main/java/org/gitlab4j/api/CommitsApi.java
@@ -1,21 +1,18 @@
 package org.gitlab4j.api;
 
+import org.gitlab4j.api.models.*;
+import org.gitlab4j.api.utils.ISO8601;
+
+import javax.ws.rs.core.Form;
+import javax.ws.rs.core.GenericType;
+import javax.ws.rs.core.Response;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.util.Date;
 import java.util.List;
 import java.util.Optional;
 
-import javax.ws.rs.core.Form;
-import javax.ws.rs.core.GenericType;
-import javax.ws.rs.core.Response;
-
-import org.gitlab4j.api.models.Comment;
-import org.gitlab4j.api.models.Commit;
-import org.gitlab4j.api.models.CommitAction;
-import org.gitlab4j.api.models.CommitPayload;
-import org.gitlab4j.api.models.Diff;
-import org.gitlab4j.api.utils.ISO8601;
+import static org.gitlab4j.api.models.CommitRef.RefType.all;
 
 /**
  * This class implements the client side API for the GitLab commits calls.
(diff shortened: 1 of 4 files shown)


src/main/java/org/gitlab4j/api/CommitsApi.java
```diff
@@ -1,21 +1,18 @@
 package org.gitlab4j.api;
 
+import org.gitlab4j.api.models.*;
```
gmessner: It is good practice to not use wildcard imports, please change this to only import the types being used.

src/main/java/org/gitlab4j/api/RepositoryApi.java
```diff
@@ -1,5 +1,13 @@
 package org.gitlab4j.api;
 
+import org.gitlab4j.api.GitLabApi.ApiVersion;
+import org.gitlab4j.api.models.*;
```
gmessner: It is good practice to not use wildcard imports, please change this to only import the types being used.

src/test/java/org/gitlab4j/api/TestCommitsApi.java
```diff
@@ -1,26 +1,20 @@
...
 import org.junit.BeforeClass;
 import org.junit.FixMethodOrder;
 import org.junit.Test;
 import org.junit.runners.MethodSorters;
 
+import javax.ws.rs.core.Response;
+import java.util.Date;
+import java.util.List;
+
+import static org.junit.Assert.*;
```
gmessner: It is good practice to not use wildcard imports, please change this to only import the types being used.

src/test/java/org/gitlab4j/api/TestCommitsApi.java
```diff
@@ -1,26 +1,20 @@
...
-import java.util.List;
-
-import javax.ws.rs.core.Response;
-
 import org.gitlab4j.api.GitLabApi.ApiVersion;
-import org.gitlab4j.api.models.Comment;
-import org.gitlab4j.api.models.Commit;
-import org.gitlab4j.api.models.Diff;
-import org.gitlab4j.api.models.Project;
+import org.gitlab4j.api.models.*;
```
gmessner: It is good practice to not use wildcard imports, please change this to only import the types being used.

src/main/java/org/gitlab4j/api/models/CommitRef.java
```diff
@@ -0,0 +1,36 @@
...
+@XmlAccessorType(XmlAccessType.FIELD)
+public class CommitRef {
+
+    private RefType type;
+    private String name;
+
+   public enum RefType {
+        branch, tag, all
+    }
+
```
gmessner: All enums should be uppercase, and to make the enum work correctly with this library this enum should look like this:
```
public enum RefType {

    BRANCH, TAG, ALL;

    private static JacksonJsonEnumHelper<RefType> enumHelper = new JacksonJsonEnumHelper<>(RefType.class);

    @JsonCreator
    public static RefType forValue(String value) {
        return enumHelper.forValue(value);
    }

    @JsonValue
    public String toValue() {
        return (enumHelper.toString(this));
    }

    @Override
    public String toString() {
        return (enumHelper.toString(this));
    }
}
```

eutkin:
ok, i done, thank you

gmessner:
@eutkin 
This has been released in 4.8.28

gmessner:
@eutkin Just a heads up.

The createRelease() and updateRelease() methods along with all the tag related methods have been moved to a new API called TagsApi.  This was done as this library tries to maintain a one-to-one relationship with the GitLab API docs, and these methods are all in the Tags API at:
https://docs.gitlab.com/ce/api/tags.html

PR #222, opened 2018-07-10
Use status code 200 instead of 201. #221

### LINKED ISSUE #221: updateIssueNote throws an GitLabApiException with the status text "OK" (status code check does not match)
# Problem
GitLab returns status code 200. But the result is matched against 201.

`api.getNotesApi().updateIssueNote(projectId, issueId, existingNoteId, "(updated): " + comment);`

`org.gitlab4j.api.GitLabApiException: OK
	at org.gitlab4j.api.AbstractApi.validate(AbstractApi.java:433) ~[gitlab4j-api-4.8.25.jar:?]
	at org.gitlab4j.api.AbstractApi.put(AbstractApi.java:303) ~[gitlab4j-api-4.8.25.jar:?]
	at org.gitlab4j.api.NotesApi.updateIssueNote(NotesApi.java:179) ~[gitlab4j-api-4.8.25.jar:?]`

# Possible fix:
Change the expected status code in **NotesApi.java:179** from 201 to 200 (Response.Status.OK)

RAYs3T:
The GitLab API returns 200 in the case of an update, not 201.
closes #221

### DIFF
--- a/src/main/java/org/gitlab4j/api/NotesApi.java
+++ b/src/main/java/org/gitlab4j/api/NotesApi.java
@@ -176,7 +176,7 @@ public class NotesApi extends AbstractApi {
         }
 
         GitLabApiForm formData = new GitLabApiForm().withParam("body", body, true);
-        Response response = put(Response.Status.CREATED, formData.asMap(), "projects", projectId, "issues", issueIid, "notes", noteId);
+        Response response = put(Response.Status.OK, formData.asMap(), "projects", projectId, "issues", issueIid, "notes", noteId);
         return (response.readEntity(Note.class));
     }
 


gmessner:
@RAYs3T 
Thanks for the fix.

gmessner:
@RAYs3T 
Just a heads up.

I did a scan of all put() calls in the project and found 2 additional places that CREATED needed to be changed to OK.  You might want to pull an update.

RAYs3T:
@gmessner 
Not sure if this is true for all the other calls since it is not documented in the GitLab API what kind of status code is returned. But i think so.

Thanks for the fast merge!

gmessner:
@RAYs3T 
Actually, it is documented here: https://docs.gitlab.com/ee/api/#status-codes

RAYs3T:
@gmessner Missed that one :3 - but well then we are fine with it.

PR #234, opened 2018-08-14
fix(systemhooks): add repositoryUpdateEvents as field to SystemHook m…

jgoldhammer:
…odel

see https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/21205

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/SystemHook.java
+++ b/src/main/java/org/gitlab4j/api/models/SystemHook.java
@@ -16,6 +16,7 @@ public class SystemHook {
     private Boolean pushEvents;
     private Boolean tagPushEvents;
     private Boolean enableSslVerification;
+    private Boolean repositoryUpdateEvents;
 
     public Integer getId() {
         return id;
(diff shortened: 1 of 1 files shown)


src/main/java/org/gitlab4j/api/models/SystemHook.java
```diff
@@ -64,4 +65,27 @@ public Boolean getEnableSslVerification() {
...
+        this.repositoryUpdateEvents = repositoryUpdateEvents;
+    }
+
+    public Boolean getRepositoryUpdateEvents() {
+        return repositoryUpdateEvents;
+    }
+
+    @Override
+    public String toString() {
+        final StringBuilder sb = new StringBuilder("SystemHook{");
```
gmessner: Prefer not to override ```toString()``` like this, these often become out of sync.  An option here would be to implement```toString()``` using something like this:
```
@Override
public String toString() {
    return (new org.gitlab4j.api.utils.JacksonJson().marshal(this));
}
```

This way the output is always in-sync with the class and outputs strings in the expected format, specifically the dates.

We could also consider adding a singleton JacksonJson instance to use for this and elsewhere.  For example, the SystemHookManager class has it's own private instance of JacksonJson, which could use the singleton instance also.

gmessner:
@jgoldhammer 
Just looking for a response on my review comment concerning toString().  I'd like to get your fix in, so please let me know what you think.  I could also just merge as is and make the toString() change afterward if you'd like.

gmessner:
@jgoldhammer 
Thank you for the fix contribution.

gmessner:
@jgoldhammer
Because Jackson is already heavily used by this library it did not require any additional dependencies, and was quite trivial to implement, it seems pretty lightweight to me.  It also provides these other advantages:

* Provides a standard approach for implementing toString() in this library
* JSON has some advantages when logging and doing log analytics
* Because it is JSON and the format can be relied on, toString()t may end up being used in ways yet to be envisioned
* Will always be synced to the properties in the class.  

Believe it or not, GitLab has often made changes and additions.  Over the life of this library, it has been one of the struggles keeping up with those changes.  Every once in a while I will review the change notes for GitLab to try and identify these.  This particular fix you've made may actually be one of those, though I have not verified that.

All the above being said, I actually prefer not to implement toString() in the model classes, and rather use the static JacksonJson.toJsonString() method I just implemented to provide that sort of thing in the calling code.

PR #236, opened 2018-08-15
fix: LAST_ACTIVITY => LAST_ACTIVITY_AT

clarkhan:
return "bad request" response `{"error":"order_by does not have a valid value"}` in GitLab version `11.1.4 `.   
fix ref: [https://docs.gitlab.com/ee/api/projects.html#list-all-projects](https://docs.gitlab.com/ee/api/projects.html#list-all-projects)


gmessner:
@clarkhan 
Thanks for the fix.

PR #237, opened 2018-08-15
Some fluent-API implementations

rubenvitt:
I've added some fluent-API getter to some models.

Factory classes are the better option in future, I think...


gmessner:
@rubenvitt 
Thanks again for the contribution.  Will let you know when the release is done.

PR #242, opened 2018-08-30
feat: update group with groupApi

rubenvitt:


gmessner:
@rubenvitt 
While you are at it, could you also add a method to add a group using a Group instance, as follows:

```
public Group addGroup(Group group) throws GitLabApiException 
```

rubenvitt:
Yes, of course. 😄

gmessner:
@rubenvitt 
Thanks for your contribution.

PR #244, opened 2018-09-04
Fix Repository Archive API Format Bug.

zhengrenjie:
If I try to visit 
` /api/v4/projects/*/repository/archive?sha=master&format=zip `,
I will get "406 Not Acceptable" error. 

Then I found this :
https://gitlab.com/gitlab-com/support-forum/issues/3067
https://gitlab.com/gitlab-org/gitlab-ce/issues/45992

This bug is caused by gitlab-ce itself, but there is a solution to request .../archive.:format instead of .../archive?format=:format. So I have changed the origin RepositoryApi.java a little to adjust to the "../archive.:format" way.

### DIFF
--- a/src/main/java/org/gitlab4j/api/RepositoryApi.java
+++ b/src/main/java/org/gitlab4j/api/RepositoryApi.java
@@ -7,8 +7,10 @@ import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
+import java.util.HashSet;
 import java.util.List;
 
+import java.util.Set;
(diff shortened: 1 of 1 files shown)


zhengrenjie:
Thank you for your best suggestions.  I have revised the PR according to your suggestion. If there is anything wrong, I will fix it as soon as I can.

gmessner:
@zhengrenjie 
Changes look good, thanks.

gmessner:
@zhengrenjie 
This has been released in 4.8.46

BTW, I added methods where you can now also specify an enum (Constants.ArchiveFormat)  for the archive format. This enum is also used to validate the format when provided as a String.

PR #245, opened 2018-09-07
Add note on Gradle requirements

etlweather:
Gradle prior to 4.5 can't pull the dependencies when running through a proxy like Nexus.


etlweather:
Good point. Fixed.

gmessner:
@etlweather 
Thanks, the changes look good, merging.

PR #246, opened 2018-09-10
Added three getIssues() methods with all optionals in Issues Api.

zhengrenjie:
Added three methods getIssues() with all optionals in Issues Api.
Added a enum IssueOrderBy.
Added a test instance.

I'm not sure if the changes were ok. If there is anything need to be fixed, I'll fix them.


gmessner:
@zhengrenjie 
Thanks for the contribution.

Going forward we want to adopt the use of filters for operations like this that have so many parameters.  See the following in ```org.gitlab4j.api.MergeRequestApi``` for an example:
```
public List<MergeRequest> getMergeRequests(MergeRequestFilter filter)
public List<MergeRequest> getMergeRequests(MergeRequestFilter filter, int page, int perPage)
public Pager<MergeRequest> getMergeRequests(MergeRequestFilter filter, int itemsPerPage)
```
The filter class is what holds all the options and provides a fluent way to easily set those options.  See ```org.gitlab4j.api.models.MergeRequestFilter``` for an example filter class.

Using the filter class allows the options to change without adding new getXXXX() methods.  Over time, GitLab tends to modify what options are used and adding new getXXXX() methods each time they do so has become a problem.

zhengrenjie:
Yeah, that will become more graceful.  And I think maybe we could create a new package in the package-models named "filter".  If you approve, I will add it then.

gmessner:
@zhengrenjie 
Creating the filter package is a good idea, though it will be a breaking change for those already using the MregeRequestFilter class.  When I release this it will become 4.9.0 because of this breaking changes.

It was determined that we will not be using the Builder pattern, instead, we simply put the withXXXX() methods in the class itself, so that the filter is constructed as follows:
```
ListIssueFilter closeFilter = new ListIssueFilter()
                .withState(IssueState.CLOSED);
```
Also, please change the class name to IssueFilter, it may actually be used for other operations other than list.

zhengrenjie:
I have made the changes. 

But I have a few questions about "getMergeRequests" method:
``` java
public List<MergeRequest> getMergeRequests(MergeRequestFilter filter, int page, int perPage) throws GitLabApiException {

        MultivaluedMap<String, String> queryParams = (filter != null ?
            filter.getQueryParams(page, perPage).asMap() : getPageQueryParams(page, perPage));
        Response response;
        if (filter != null && (filter.getProjectId() != null && filter.getProjectId().intValue() > 0) ||
                (filter.getIids() != null && filter.getIids().size() > 0)) {

            if (filter.getProjectId() == null || filter.getProjectId().intValue() == 0) {
                throw new RuntimeException("project ID cannot be null or 0");
            }

            response = get(Response.Status.OK, queryParams, "projects", filter.getProjectId(), "merge_requests");
        } else {
            response = get(Response.Status.OK, queryParams, "merge_requests");
        }

        return (response.readEntity(new GenericType<List<MergeRequest>>() {}));
    }
```
First of all, I think the "projectIdOrPath" parameter should not put into MergeRequestFilter, for projectIdOrPath is actually required  by gitlab api. Base on my guess, this method try to combine two api which need projectIdOrPath parameter and not into one method, and split them in the method by test whether filter.getProjectId() equals null. I think that is not a good design. 
Second, testing if filter equals null may not be necessary, because if I don't want to add some filters about query, I should use other method without this field instead of passing a null.

What above is just my idea.  ^_^.

I am going to fix it. But I found there are apis could not use the same "IssueFilter" except listing apis, so 
I name it with "ListIssueFilter". And still, I will change "ListIssueFilter" to "IssueFilter".

zhengrenjie:
I have finished. ^_^
Please trouble check it again.

gmessner:
There is nothing stopping anyone calling the method with a null or empty filter, when it is null it will simply return all items.  

The methods that use the MergeRequestFilter automatically determine which endpoint to call based on whether the project ID is present in the filter.  If we did it otherwise, we would need to create 2 different versions of the getMergeRequests() methods as you have done with getIssues().  I believe that having a separate method with and without the project specified is actually not a good idea.  Basically a single field that you can filter on has been pulled out of the filter and added as a parameter to the method.

Concerning the use of the "List" prefix.  There may not currently be any methods other than listing issues that use IssueFilter, but you never know.  Also, prefixing the issue filter or any other filter with the word "List" does not provide any clarity on its purpose, so it is not needed.  I will be enforcing this pattern for all filters created in the future.

gmessner:
@zhengrenjie 
An issue was just reported concerning the labels parameter on the MergeRequestFilter (#250).  Basically, the GitLab API is wanting a comma-separated list instead of a query array, I just looked at the API docs and get issues is the same way.  You'll need to change IssueFilter.getQueryParams() as follows:
```
.withParam("labels", (labels != null ? String.join(",", labels) : null))
```
I'll merge once this is done

zhengrenjie:
Finished !

PR #252, opened 2018-09-13
feat-impl: custom attributes for users

rubenvitt:
... can't be tested well, so please double-check. I'll try to fix, if you find something.

### DIFF
--- a/src/main/java/org/gitlab4j/api/AbstractApi.java
+++ b/src/main/java/org/gitlab4j/api/AbstractApi.java
@@ -113,13 +113,13 @@ public abstract class AbstractApi implements Constants {
     protected GitLabApiClient getApiClient() {
         return (gitLabApi.getApiClient());
     }
-    
+
     /**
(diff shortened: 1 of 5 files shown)


gmessner:
Haven't tested yet, but I believe this may fail if the authenticated user is not an administrator. https://docs.gitlab.com/ee/api/custom_attributes.html states the following:

"Every API call to custom attributes must be authenticated as administrator."

Also, I'm not a fan of adding ```"?with_custom_attributes=true"``` as part of the path, it should be moved to the query params that are being passed in. 

I'm thinking it might be better to make custom attributes a property of the UserApi so that one could enable custom attributes by doing something like this:
```
gitLabApi.getUserApi().enableCustomAttributes();
List<User> users = gitLabApi.getUserApi().getUsers();
```

Or on a per call basis like this:
```
List<User> users = gitLabApi.getUserApi().withCustomAttributes().getUsers();
```

rubenvitt:
I think, that's all. :)
Thanks!

PR #255, opened 2018-09-20
feat: changing customAttributes for users

rubenvitt:
See #252 - api-users should be able to change customAttributes for users.


gmessner:
@rubenvitt 
Thanks for the contribution.

It feels like the methods are backward for all of the new methods.  The methods that take String parameters should be the ones making the calls to the server, instead of creating a CustomAttribute instance and passing that to the other method that then pulls the Strings out of the CustomAttribute to pass to the server.   Doing it this way will also verify the key and value of the CustomAttribute which is not currently being done.

rubenvitt:
Should be okay now.

gmessner:
@rubenvitt 
Looks good.  And thanks again for the contribution.

PR #262, opened 2018-10-26
Use correct visibility value in addGroup(Group)

mdeknowis:
The implementation `org.gitlab4j.api.GroupApi.addGroup(Group)` has in small bug:

The parameter `visibility` is filled with the value of `group.getDescription()` instead of  `group.getVisibility()`

Hence the request is failing, if description is set. Also it's not possible to change the default visibility `private`.


gmessner:
@mdeknowis 
Once again, thank you for your contribution.

mdeknowis:
@gmessner Thanks!

PR #263, opened 2018-10-26
Add support for resolve_outdated_diff_discussions in Project

mdeknowis:
The [Projects API](https://docs.gitlab.com/ee/api/projects.html) supports `resolve_outdated_diff_discussions`, which is missing in the current implementation.

So adding support for `resolve_outdated_diff_discussions` to `Project`, `ProjectApi.createProject` and `ProjectApi.updateProject` allows API user to adjust the value.

Also adding support for `initialize_with_readme` to to `Project` and `ProjectApi.createProject`


mdeknowis:
@gmessner Thanks!

PR #269, opened 2018-11-05
Add all parameters to get project of group api

mdeknowis:
[List a group’s projects](https://docs.gitlab.com/ee/api/groups.html#list-a-groups-projects) provides a wide API to filter for projects of a group. That would help to limit the result.


src/main/java/org/gitlab4j/api/Constants.java
```diff
@@ -199,6 +199,28 @@ public String toString() {
         }
     }
 
+    /** Enum to use for ordering the results of getProjects() of GroupsApi. */
+    public enum GroupProjectOrderBy {
+
```
gmessner: There is no need for this enum, the same enums are defined in ProjectOrderBy.

src/main/java/org/gitlab4j/api/GroupApi.java
```diff
@@ -224,6 +224,46 @@ public GroupApi(GitLabApi gitLabApi) {
...
+          .withParam("search", search)
+          .withParam("simple", simple)
+          .withParam("owned", owned)
+          .withParam("starred", starred)
+          .withParam("with_issues_enabled", withIssuesEnabled)
+          .withParam("with_merge_requests_enabled", withMergeRequestsEnabled)
+          .withParam("with_custom_attributes", withCustomAttributes);
+        Response response = get(Response.Status.OK, formData.asMap(), "groups", groupId, "projects");
+        return (response.readEntity(new GenericType<List<Project>>() {}));
+    }
```
gmessner: For all new methods that take this many properties, this library is now using the Filter pattern.  See the ```org.gitlab4j.api.models.ProjectFilter``` and its use in the following methods in the ProjectApi:
```
public List<Project> getUserProjects(Object userIdOrUsername, ProjectFilter filter)
public List<Project> getUserProjects(Object userIdOrUsername, ProjectFilter filter, int page, int perPage)
public Pager<Project> getUserProjects(Object userIdOrUsername, ProjectFilter filter, int itemsPerPage)
```

src/main/java/org/gitlab4j/api/Constants.java
mdeknowis: Oh, thanks. I will remove it.

src/main/java/org/gitlab4j/api/GroupApi.java
mdeknowis: Oh, that's better!

mdeknowis:
@gmessner Thanks for your feedback. I hope, I adjusted the code in a manner, that fits to your project.

src/main/java/org/gitlab4j/api/GroupApi.java
mdeknowis: Did you consider to use [Lombok](https://projectlombok.org/) ? Then you could simply get the `withXXX` for free by using [@Setter](https://projectlombok.org/features/GetterSetter)?

src/main/java/org/gitlab4j/api/GroupApi.java
```diff
@@ -224,6 +225,32 @@ public GroupApi(GitLabApi gitLabApi) {
         return (new Pager<Group>(this, Group.class, itemsPerPage, formData.asMap(), "groups", groupId, "subgroups"));
     }
 
+    /**
+     * Get a list of projects belonging to the specified group ID.
+     *
+     * GET /groups/:id/projects
+     *
+     * @param groupId the group ID to list the projects for
```
gmessner: The below ```@params``` need to be removed or an error occurs when the Javadocs are generated.

src/main/java/org/gitlab4j/api/GroupApi.java
```diff
@@ -224,6 +225,32 @@ public GroupApi(GitLabApi gitLabApi) {
...
+     * @param simple Return only the ID, URL, name, and path of each project
+     * @param owned Limit by projects owned by the current user
+     * @param starred Limit by projects starred by the current user
+     * @param withIssuesEnabled Limit by enabled issues feature
+     * @param withMergeRequestsEnabled Limit by enabled merge requests feature
+     * @param withCustomAttributes Include custom attributes in response (admins only)
+     * @return a list of projects belonging to the specified group ID
+     * @throws GitLabApiException if any exception occurs
+     */
+  public List<Project> getProjects(int groupId, ProjectOfGroupFilter filter) throws GitLabApiException {
```
gmessner: For new methods that take a group ID this library is now using ```Object groupIdOrPath``` instead of ```int groupId```.
You then use ```getGroupIdOrPath()``` like this:
```
    Response response = get(Response.Status.OK, formData.asMap(), "groups",  getGroupIdOrPath(groupIdOrPath), "projects");
    return (response.readEntity(new GenericType<List<Project>>() {}));
```

src/main/java/org/gitlab4j/api/GroupApi.java
gmessner: Lombok is great, however for a library like this,  it is better to be explicit, specifically for documentation (Javadoc)  purposes. 

We have experimented using Lombok, but getting the serialization/deserialization correct can also be problematic, especially with a library of this size.

mdeknowis: Oh okay... Interesting to know.

src/main/java/org/gitlab4j/api/GroupApi.java
mdeknowis: Oh sorry, forgot to keep the doc in sync.

src/main/java/org/gitlab4j/api/GroupApi.java
mdeknowis: Adjusted. 👍

mdeknowis:
@gmessner So please don't hesitate, if I still missed something, and didn't fit to our project's style and architecture.

src/main/java/org/gitlab4j/api/models/ProjectOfGroupFilter.java
```diff
@@ -0,0 +1,167 @@
...
+import org.gitlab4j.api.Constants.ProjectOrderBy;
+import org.gitlab4j.api.Constants.SortOrder;
+import org.gitlab4j.api.Constants;
+import org.gitlab4j.api.GitLabApiForm;
+
+/**
+ *  This class is used to filter Projects when getting lists of projects for a specified group.
+ */
+public class ProjectOfGroupFilter {
+
```
gmessner: @mdeknowis 
I should have mentioned this in an earlier comment, but simply forgot.  The name of this class would be more descriptive as ```GroupProjectFilter```.  I am going to go ahead and merge, but change the name during the release process.

gmessner:
@mdeknowis 
The changes look great.  I'm going to merge this. However, I will be renaming the ```ProjectOfGroupFilter``` to ```GroupProjectsFilter``` as I prepare it for release.  Could of requested this changed but wanted to get this merged so it could be used.

Thanks again for your contributions.

gmessner:
4.8.55 has been released with your enhancement.

mdeknowis:
Thanks a lot!

Issue #271, opened 2018-11-13
Question: Missing equals() and hashCode() implementations


mdeknowis:
Hello,

we stumbled across the fact none of the DTOs are implementing `equals` and `hashCode` (and `toString`).

Currently we are forced to use `org.apache.commons.lang3.builder.ToStringBuilder.reflectionToString(Object)` and `org.mockito.ArgumentMatchers.refEq(T, String...)` etc. to test our code and to generate log output.

So we wonder, if there is a particular reason for that fact?

Regards

gmessner:
@mdeknowis 
For these sorts of classes that are serializing and deserializing JSON toString() methods are not that useful, we provided ```org.gitlab4j.api.utils.JacksonJson.toJsonString()``` so you can simply do this:
```
String s = org.gitlab4j.api.utils.JacksonJson.toJsonString(obj)
```

For comparison, I suggest you look at how we do it in our unit tests, specifically TestGitLabApiBeans.java
I have been thinking about generalizing the comparison code from this and putting it in JacksonJson, if you believe that is a good idea, let me know.

mdeknowis:
Thanks for that `org.gitlab4j.api.utils.JacksonJson.toJsonString()` hint. Currently we try avoid to build `if (log.isDebugEnabled()) {...}` blocks, so we rely on `toString()`. Also in our test we want still to use the advantages of `AbstractObjectAssert` e.g. `org.assertj.core.api.AbstractObjectAssert.isEqualToComparingFieldByFieldRecursively(Object)`, which has more detailed output regarding actual and expected.

mdeknowis:
The issue is also, that frameworks like Mockito use `toString`:

```
Argument(s) are different! Wanted:
org.gitlab4j.api.ProjectApi#0 bean.createProject(
    refEq(org.gitlab4j.api.models.Project@7495699f)
);
-> at my.GitlabProjectServiceTest.testAddProject(GitlabProjectServiceTest.java:202)
Actual invocation has different arguments:
org.gitlab4j.api.ProjectApi#0 bean.createProject(
    org.gitlab4j.api.models.Project@3047254d
);
-> at my.GitlabProjectService.addProject(GitlabProjectService.java:118)

	at my.GitlabProjectServiceTest.testAddProject(GitlabProjectServiceTest.java:202)
...
```

mdeknowis:
Also during debugging, it would be more developer friendly:
[...]

PR #273, opened 2018-11-13
Add withXxxxx() methods to AbstractUser

mdeknowis:
As discussed in https://github.com/gmessner/gitlab4j-api/issues/272 adding the missing `with*` methods to `AbstractUsers`


gmessner:
@mdeknowis 
Reviewing this pull request made me realize 2 things:
1) AbstractUser is not marked abstract, and it should be.
2) the ```withXxxx()``` methods do not work on a base class.  The ```withXxxx()``` methods are meant to be chained together, which will not work on the sub-classes of AbstractUser.  They can be chained but cannot be assigned to a sub-class without casting.  This is why the ```User``` class has implemented the ```withXxxx()``` methods.

An example trying to use with a sub-class:

```
Assignee assignee = new Assignee()
    .withAvatarUrl("https://asdad.com/qqwwe")
    .withBio("https://wqeqwqew.com/qweqwe");
```
Will not compile unless you add a type cast as follows:
```
Assignee assignee = (Assignee) new Assignee()
    .withAvatarUrl("https://asdad.com/qqwwe")
    .withBio("https://wqeqwqew.com/qweqwe");
```

What sub-class were you trying to use when you needed the ```withXxxx()``` methods?

gmessner:
@mdeknowis 
I've put some more thought into this, and there is a way to fix this.  Need to make AbstractUser a generic type, i.e.:
```
public abstract class AbstractUser<T>
```
Change all the ```withXxxx()``` methods in AbstractUser to look like:
```
public AbstractUser<T> withAvatarUrl(String avatarUrl)
```

And update all the sub-classes as follows:
```
public class Assignee extends AbstractUser<Assignee>
public class Author extends AbstractUser<Author>
public class Contributor extends AbstractUser<Contributor>
public class Participant extends AbstractUser<Participant>
public class User extends AbstractUser<User>
```
Once that is done, the ```User``` class can be updated to remove all the duplicate ```withXxxx()``` methods.

mdeknowis:
@gmessner I adjusted the pull request according to your proposal

src/main/java/org/gitlab4j/api/models/User.java
```diff
@@ -15,98 +14,27 @@ public String getExternUid() {
...
-        setOrganization(organization);
-        return this;
-    }
-
+    /**
+     * @deprecated Replaced by {@link #withProjectsLimit(Integer)}
+     * @see #withProjectsLimit(Integer) 
+     */
+    @Deprecated
     public User withProjectLimit(Integer projectsLimit) {
```
gmessner: Is there a reason to not remove this like the other with* methods that were removed?

src/main/java/org/gitlab4j/api/models/User.java
```diff
@@ -15,98 +14,27 @@ public String getExternUid() {
...
-        setExternal(external);
-        return this;
-    }
-
+    /**
+     * @deprecated Replaced by {@link #withSharedRunnersMinutesLimit(Integer)}
+     * @see #withSharedRunnersMinutesLimit(Integer) 
+     */
+    @Deprecated
     public User withSharedRunnersMinuteLimit(Integer sharedRunnersMinuteLimit) {
```
gmessner: Is there a reason to not remove this like the other with* methods that were removed?

gmessner:
@mdeknowis 
BTW, As soon as we get this merged, I will also merge the changes that add toString() implementations as per Issue #271

src/main/java/org/gitlab4j/api/models/User.java
mdeknowis: @gmessner I didn't want to break the API. I don't know, if someone uses that method. But I can also remove these methods instead of deprecating them, if you like.

src/main/java/org/gitlab4j/api/models/User.java
mdeknowis: @gmessner I didn't want to break the API. I don't know, if someone uses that method. But I can also remove these methods instead of deprecating them, if you like.

src/main/java/org/gitlab4j/api/models/User.java
gmessner: @mdeknowis 
I got it, I missed the fact that an "s" was missing. Leave it in as deprecated.

src/main/java/org/gitlab4j/api/models/User.java
gmessner: @mdeknowis 
I got it, I missed the fact that an "s" was missing. Leave it in as deprecated.

gmessner:
@mdeknowis 
Thanks again for the contribution.  I'll let you know when it is released.

PR #279, opened 2018-12-10
Add Discussions model and api

filippobuletto:
Discussions are set of related notes on snippets, issues, epics, merge requests or commits.

This commit only include merge requests one.

### DIFF
--- a/src/main/java/org/gitlab4j/api/MergeRequestApi.java
+++ b/src/main/java/org/gitlab4j/api/MergeRequestApi.java
@@ -10,6 +10,7 @@ import javax.ws.rs.core.Response;
 
 import org.gitlab4j.api.GitLabApi.ApiVersion;
 import org.gitlab4j.api.models.Commit;
+import org.gitlab4j.api.models.Discussion;
 import org.gitlab4j.api.models.Issue;
 import org.gitlab4j.api.models.MergeRequest;
 import org.gitlab4j.api.models.MergeRequestFilter;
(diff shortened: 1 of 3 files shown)


src/main/java/org/gitlab4j/api/MergeRequestApi.java
```diff
@@ -279,6 +280,44 @@ public MergeRequest getMergeRequest(Integer projectId, Integer mergeRequestIid)
                 "projects", projectId, "merge_requests", mergeRequestIid, "commits"));
     }
 
+    /**
```
gmessner: The discussions related methods should be moved to a class called ```DiscussionApi```.  This library follows a pattern where we add an API class for each API described in the GitLab API documentation.  This one is the "Discussions API" as described here: https://docs.gitlab.com/ee/api/discussions.html

src/main/java/org/gitlab4j/api/models/Discussion.java
```diff
@@ -0,0 +1,48 @@
...
+import javax.xml.bind.annotation.XmlRootElement;
+
+import org.gitlab4j.api.utils.JacksonJson;
+
+@XmlRootElement
+@XmlAccessorType(XmlAccessType.FIELD)
+public class Discussion {
+
+  private String id;
+  private String individualNote;
```
gmessner: This should be a Boolean, not a String.  Will need to update the set and get methods also.

src/main/java/org/gitlab4j/api/models/Note.java
```diff
@@ -74,6 +74,9 @@ public String toString() {
     private String title;
     private String updatedAt;
     private Boolean upvote;
+    private Boolean resolved;
+    private Boolean resolvable;
+    private Participant resolvedBy;
```
gmessner: ```type`` is also missing from the Note properties.  This should be an Enum, with the current values being "DiscussionNote" and "DiffNote"

gmessner:
@filippobuletto 
If you would prefer, I can go ahead and merge this and make the requested changes myself.  I will also add the other methods from the Discussions API when doing so.

src/main/java/org/gitlab4j/api/models/Note.java
filippobuletto: I cannot find the model description but I'm changing the class according to your comment!

filippobuletto:
It's OK to me if you want to merge and complete the changes! Thank you for the suggestions, meanwhile I'll stick to my local changes until you complete the missing api 👍

src/main/java/org/gitlab4j/api/models/Note.java
gmessner: You'll find a sample response here:
https://docs.gitlab.com/ee/api/discussions.html#list-project-issue-discussions

Notice the items in the "notes" array have a type property.  Thanks for adding this.

gmessner:
@filippobuletto 
Just realized that ```GitLabApi``` had not been modified to supply the ```DiscussionsApi``` instance.  This has been fixed and released in 4.9.5.  Example use:
```
List<Discussion> mrDiscussions = gitLabApi.getDiscussionsApi().getMergeRequestDiscussions(projectId, mergeRequestIid);
```

PR #293, opened 2019-02-02
chmod commit action

augi:
GitLab API now [supports](https://docs.gitlab.com/ce/api/commits.html#create-a-commit-with-multiple-files-and-actions) also `CHMOD` commit action, and `execute_filemode` parameter (`boolean`).


gmessner:
@augi 
Thanks for your contribution.

PR #297, opened 2019-02-10
Spliterator improvments

mariuszs:
- fix for NPE on empty source
- add NPE when action is missing (required by contract)
- declare characteristics


gmessner:
@mariuszs 
Thanks again for your contributions.

PR #298, opened 2019-02-10
Java 8 Stream Documentation improvements

gmessner:
Clarifies the use of Streams for both eager and lazy evaluation uses.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ To utilize the GitLab API for Java in your project, simply add the following dep
 ```java
 dependencies {
     ...
-    compile group: 'org.gitlab4j', name: 'gitlab4j-api', version: '4.9.15'
+    compile group: 'org.gitlab4j', name: 'gitlab4j-api', version: '4.9.16'
 }
 ```
 
@@ -22,7 +22,7 @@ dependencies {
 <dependency>
     <groupId>org.gitlab4j</groupId>
     <artifactId>gitlab4j-api</artifactId>
-    <version>4.9.15</version>
+    <version>4.9.16</version>
 </dependency>
 ```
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -141,34 +141,44 @@ List<Project> allProjects = projectPager.all();
...
-```
-GroupApi
-IssuesApi
-NotesApi
-ProjectApi
-RepositoryApi
-TagsApi
-UserApi
+// Stream method
+Stream projectsStream<Project> = gitlabApi.getProjectApi().getProjectsStream();
```
mariuszs: duplicated Stream<Project> at the beginning

README.md
```diff
@@ -141,34 +141,44 @@ List<Project> allProjects = projectPager.all();
...
 ```
+
 Example usage:
+
 ```java
 // Stream the visible Projects printing out the project name.
-gitlabApi.getProjectsApi().getProjectsStream().map(Project::getName).forEach(name -> System.out.println(name));
+Stream projectsStream<Project> = gitlabApi.getProjectApi().getProjectsStream();
```
mariuszs: Stream<Project> projects

README.md
```diff
@@ -141,34 +141,44 @@ List<Project> allProjects = projectPager.all();
...
+projectsStream.map(Project::getName).forEach(name -> System.out.println(name));
 
 // Operate on the stream in parallel, this example sorts User instances by username
```
mariuszs: I'm not sure if this is a good idea to give such an example here. Parallel operations are generally used for performance optimization, but here project gathering is so slow, that this operation doesn't add any improvement. As a result, the user can be confused that this operation can be super fast and this library can fetch data in multiple channels.

src/main/java/org/gitlab4j/api/Pager.java
```diff
@@ -312,34 +313,62 @@ public void remove() {
...
+     * @return a Stream instance for lazily streaming items from the GitLab server
+     * @throws IllegalStateException if Stream has already been issued
+     */
+    public Stream<T> lazyStream() throws IllegalStateException {
+
+        if (pagerStream == null) {
+            synchronized (this) {
+                if (pagerStream == null) {
+
+                }
```
mariuszs: wrong brace position

src/test/java/org/gitlab4j/api/TestStreams.java
```diff
@@ -85,4 +85,35 @@ public void testParallelStream() throws Exception {
...
+            assertTrue(usernames.contains(sortedUsers.get(i).getUsername()));
+        }
+    }
+
+    @Test
+    public void testStreamLimit() throws Exception {
+
+        // Arrange
+        Stream<User> stream = new UserApi(gitLabApi).getUsersStream();
+
```
mariuszs: This test has sense only for lazy streams, for normal stream it is guaranteed by Java API that this always pass, so an additional test is not needed. But for the lazy stream, it will be nice to add such test, with pagination lesser than a limit, because we need to be sure, that our splitter can gather data from gitlab.

README.md
```diff
@@ -141,34 +141,44 @@ List<Project> allProjects = projectPager.all();
...
 Stream<User> stream = new UserApi(gitLabApi).getUsersStream();
-List<User> sortedUsers = stream.parallel().sorted(comparing(User::getUsername)).collect(toList());
+List<User> users = stream.parallel().sorted(comparing(User::getUsername)).collect(toList());
+```
+
+**IMPORTANT**  
+The built-in methods that return a Stream do so using ___eager evaluation___, meaning all items are pre-fetched from the GitLab server and a Stream is returned which will stream those instances.  This allows for paralell processing of the Stream.
+
+To stream using ___lazy evaluation___, utilize the GitLab4J-API methods that return a ```Pager``` instance, and then call the ```lazyStream()``` method on the ```Pager``` instance to create a lazy evaluation Stream.
```
mariuszs: I will add information about how it works: Lazy stream is fetching data page by page and can stop earlier when a limit is specified on stream. Fetching always starts from the beginning, so it cannot skip elements. This stream doesn't support parallel operations.

src/main/java/org/gitlab4j/api/Pager.java
gmessner: Fixed

README.md
gmessner: I added some info regarding lazy streams with the latest commits.

README.md
gmessner: Fixed.

README.md
gmessner: Fixed.

src/test/java/org/gitlab4j/api/TestStreams.java
gmessner: Good point, updated.

README.md
gmessner: Parallel operations are often performed after fetching the data, paralell operations are not only used to speed up fetching, I have added more info to make the fact that reading of the data for the Stream cannot done in paralell.

gmessner:
@mariuszs 
Thanks for all your input.

PR #299, opened 2019-02-10
verify that PagerSplitter can collect data page by page

mariuszs:


gmessner:
@mariuszs 
Thanks again for the contribution.

Issue #291, opened 2019-01-22
Create discussions


tomasbjerre:
I was looking for a way to create new discussions. And specifically on merge requests. https://docs.gitlab.com/ee/api/discussions.html#create-new-merge-request-discussion

This is not currently supported right?

gmessner:
@tomasbjerre 
Will will include discussion management methods to the DiscussionsApi in the next release scheduled for 1/28/2019

gmessner:
@tomasbjerre 
My apologies for taking so long to get this released.  The DiscussionsApi now has the ability to create and resolve merge request discussions and delete notes from a merge request discussion.

4.9.15 is now available, please verify and close this issue when you have time.

tomasbjerre:
Thanks!!

PR #300, opened 2019-02-11
Adding diff_refs in MergeRequest #291

tomasbjerre:
Need this as input to Discussions API.

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/DiffRef.java
+++ b/src/main/java/org/gitlab4j/api/models/DiffRef.java
@@ -1,0 +1,36 @@
+package org.gitlab4j.api.models;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class DiffRef {
+    private String baseSha;
+    private String headSha;
+    private String startSha;
+
+    public DiffRef() {
+    }
+
+    public String getBaseSha() {
+        return baseSha;
+    }
+
+    public void setBaseSha(final String baseSha) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/gitlab4j/api/models/DiffRef.java
```diff
@@ -0,0 +1,34 @@
+package org.gitlab4j.api.models;
+
+public class DiffRef {
+    private String base_sha;
+    private String head_sha;
+    private String start_sha;
+
```
gmessner: This library utilizes camel-case for variable names, so these should be:
```
    private String baseSha;
    private String headSha;
    private String startSha;
```

As should the getters and setters for same.

src/main/java/org/gitlab4j/api/models/MergeRequest.java
```diff
@@ -47,6 +45,7 @@
     private Integer userNotesCount;
     private String webUrl;
     private Boolean workInProgress;
+    private DiffRef diff_refs;
```
gmessner: Use camel-case:
```
 private DiffRef diffRefs
```

Same goes for the getters and setters.

tomasbjerre:
Resolved

tomasbjerre:
@gmessner issues resolved.

gmessner:
@tomasbjerre 
Thank you for your contribution

tomasbjerre:
A new release with this included would be great!

PR #304, opened 2019-02-15
Added Group Milestones API

### LINKED ISSUE #303: Improvement: missing start_date attribute in Milestone model
A GitLab Milestone has a `start_date` attribute that is missing in the `gitlab4j-api` model.

See: https://docs.gitlab.com/ee/api/milestones.html#list-project-milestones

pismy:
see: https://docs.gitlab.com/ee/api/group_milestones.html

closes #303

### DIFF
--- a/src/main/java/org/gitlab4j/api/MilestonesApi.java
+++ b/src/main/java/org/gitlab4j/api/MilestonesApi.java
@@ -22,6 +22,240 @@ public class MilestonesApi extends AbstractApi {
     }
 
     /**
+     * Get a list of group milestones.
+     *
+     * @param groupIdOrPath the group in the form of an Integer(ID), String(path), or Group instance
+     * @return the milestones associated with the specified group
+     * @throws GitLabApiException if any exception occurs
+     */
+    public List<Milestone> getGroupMilestones(Object groupIdOrPath) throws GitLabApiException {
+        return (getGroupMilestones(groupIdOrPath, getDefaultPerPage()).all());
+    }
+
+    /**
+     * Get a list of group milestones.
+     *
+     * @param groupIdOrPath the group in the form of an Integer(ID), String(path), or Group instance
+     * @param page the page number to get
+     * @param perPage how many milestones per page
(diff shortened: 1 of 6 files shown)


src/main/java/org/gitlab4j/api/MilestonesApi.java
```diff
@@ -21,6 +21,240 @@ public MilestonesApi(GitLabApi gitLabApi) {
...
+
+    /**
+     * Get the specified group milestone.
+     *
+     * @param groupIdOrPath the group in the form of an Integer(ID), String(path), or Group instance
+     * @param milestoneId the ID of the milestone tp get
+     * @return a Milestone instance for the specified IDs
+     * @throws GitLabApiException if any exception occurs
+     */
+    public Milestone getGroupMilestone(Object groupIdOrPath, int milestoneId) throws GitLabApiException {
```
gmessner: For consistency, ```int``` should be ```Integer```

pismy: Done.

Just for your knowledge, the exiting [getMilestone()](https://github.com/gmessner/gitlab4j-api/blob/master/src/main/java/org/gitlab4j/api/MilestonesApi.java#L131) method also has the wrong signature.
I didn't change it as I don't manage the impacts.

gmessner: @pismy 
I will change the existing ```getMilestone()``` post-merge, unless you'd like to make the change.  It should have no impact.

pismy: just did it.

pismy: How often do you make releases ?

gmessner: @pismy 
I typically try and release when a PR is merged, and will usually include updates that are a result of issues being fixed or implemented.  SO you can expect a release either today or tomorrow.

PR #319, opened 2019-03-11
Update with pipeline schedule support

lpiet:
Hopefully doing this the right way, but I figured if I open the issue I might contribute to the solution. 

Tested it against our on premise gitlab instance and it seems to work. Hopefully should help implement the support for the pipeline schedules.


src/main/java/org/gitlab4j/api/ProjectApi.java
```diff
@@ -2395,4 +2395,166 @@ public Project setProjectAvatar(Object projectIdOrPath, File avatarFile) throws
         Response response = putUpload(Response.Status.OK, "avatar", avatarFile, "projects", getProjectIdOrPath(projectIdOrPath));
         return (response.readEntity(Project.class));
     }
```
gmessner: The pipeline schedule methods should be moved to the ```org.gitlab4j.api.PipelineApi``` class

lpiet:
@gmessner,

I believe this covers it. I did notice that if I copy the exact json from the site any timestamp with a millisecond field of 000 fails the test (if it has any millisecond value it works) . But this was also the case for the existing tests. 

The json converter strips the trailing 0's and then the generated and input json don't match.

ex. "next_run_at": "2017-05-19T13:41:00.000Z".

PS
I'll have one or two extra pull requests up next week, working through the functions I need for the role out at work and adding where needed.

gmessner:
@lpiet 
There are caveats when testing the JSON, this being one of them, I should have mentioned this, my apologies if you spent much time on this.  The other big caveat is that values in the test file cannot have ```null``` as a value.

gmessner:
@lpiet 
I am assuming that you are done with this PR?

lpiet:
correct

gmessner:
@lpiet 
This has been merged, A release will happen later this week.  WIll let you know when the release is completed.

PR #360, opened 2019-05-20
Support event name on web hooks

PhilippeVienne:
Type of PR: Bug resolution
Kind of bug: Null pointer on web hook event types
Solution: https://stackoverflow.com/questions/33611199/jackson-jsontypeinfo-property-is-being-mapped-as-null

It allows the following usage:
```java
    @Override
    public void onProjectEvent(ProjectSystemHookEvent event) {
        switch (event.getEventName()) {
            case ProjectSystemHookEvent.PROJECT_CREATE_EVENT:
                log.info("Project created {}", event.getName());
                break;
            case ProjectSystemHookEvent.PROJECT_DESTROY_EVENT:
                log.info("Project destroyed {}", event.getName());
                break;
        }
    }
```

### DIFF
--- a/src/main/java/org/gitlab4j/api/systemhooks/SystemHookEvent.java
+++ b/src/main/java/org/gitlab4j/api/systemhooks/SystemHookEvent.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 
 @JsonTypeInfo(use=JsonTypeInfo.Id.NAME,
+    visible = true,
     property="event_name")
 @JsonSubTypes({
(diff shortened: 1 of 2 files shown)


gmessner:
@PhilippeVienne 
Nice find. I have merged the PR, will include on the next release which should be in the next 1-2 days.

I have to say Jackson sure picked an unusual way to do this, not passing a field down to the deserializer by default because it was used in type selection seems a bit odd, It does indicate this as the behavior if you thoroughly read the Javadocs on @JsonTypeInfo (https://fasterxml.github.io/jackson-annotations/javadoc/2.9/com/fasterxml/jackson/annotation/JsonTypeInfo.html), they should consider making visible ```true``` by default, as many people have run into this problem.

Issue #367, opened 2019-06-06
Container Registry API is not implemented


PhilippeVienne:
The Container Registry API mentioned https://docs.gitlab.com/ee/api/container_registry.html is not yet implemented in GitLab4J.

I will try to have a look to implement it.

PR #368, opened 2019-06-06
Add Container Registry API

PhilippeVienne:
Pull request to resolve: #367 

- [x] Implement Container Registry API following documentation
- [ ] Add Documentation about API on readme
- [ ] Implement Testing for the API

### DIFF
--- a/src/main/java/org/gitlab4j/api/ContainerRegistryApi.java
+++ b/src/main/java/org/gitlab4j/api/ContainerRegistryApi.java
@@ -1,0 +1,227 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Greg Messner <greg@messners.com>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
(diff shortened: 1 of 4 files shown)


src/main/java/org/gitlab4j/api/ContainerRegistryApi.java
```diff
@@ -0,0 +1,227 @@
...
+     * @param olderThan       Tags to delete that are older than the given time, written in human readable form
+     *                        <code>1h</code>, <code>1d</code>, <code>1month</code>.
+     * @throws GitLabApiException if any exception occurs
+     */
+    @SuppressWarnings({"RedundantThrows", "unused"})
+    public void deleteRepositoryTagBulk(Object projectIdOrPath, Integer repositoryId, String nameRegex, Integer keepN, String olderThan) throws GitLabApiException {
+        /*
+            This function is not implemented because we need the library to support FormData on DELETE requests.
+            See: https://docs.gitlab.com/ee/api/container_registry.html#delete-repository-tags-in-bulk
+         */
```
gmessner: It is highly unusual to have form data with the DELETE method.  Have you tried this with query params?

PhilippeVienne: I was just following the curl request set as example on the GitLab documentation. I will have a try with params.

gmessner: @PhilippeVienne 
Let me know how it goes.  If it requires form data, I'll add a ````delete()``` method that accepts form data.

PhilippeVienne: My local test have correctly a 202 Accepted response but the bulk DELETE seems to fail in all situation, as stated by documentation and also as you suggested with Query Params

gmessner:
@PhilippeVienne 
I forgot to thank you for your contribution, so thank you.

I am going to go ahead and merge this PR.  Once merged, I'll look into and implement the bulk delete functionality.  I will also create the ```TestContainerRegistryApi``` test class but would like to work with you to fully implement the integration tests for this. Will provide more information later today.

gmessner:
@PhilippeVienne 
The bulk tag delete is working correctly using query parameters, what I did find out is that if you call the bulk delete API endpoint more than once in an hour, the GitLab server simply responds with a 202 status and does absolutely nothing.  It does work on the first use though or done less than once an hour.  The “at most once an hour” is mentioned in the GitLab documentation (https://docs.gitlab.com/ee/api/container_registry.html#delete-repository-tags-in-bulk)

PR #372, opened 2019-06-06
Added encoding to the tag name where the tag name is used in the url path

### LINKED ISSUE #369: Tag api should encode tag name
I think the tag name might need to be url encoded.  If you make the call to get a tag with a tag named something like the following: `env/STAGING` you will get a 404 even if the tag exists.  I think this is due to the slash and the fact that in the API the tag name goes on the url itself.

chrisrohr:
The GET and DELETE calls on Gitlab put the tag name on the path itself.  This is problematic if you have a tag with URL reserved characters (i.e. a slash).  This should fix #369 by url encoding the tag prior to setting it on the path.

### DIFF
--- a/src/main/java/org/gitlab4j/api/TagsApi.java
+++ b/src/main/java/org/gitlab4j/api/TagsApi.java
@@ -94,7 +94,7 @@ public class TagsApi extends AbstractApi {
      * @throws GitLabApiException if any exception occurs
      */
     public Tag getTag(Object projectIdOrPath, String tagName) throws GitLabApiException {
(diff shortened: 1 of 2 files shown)


gmessner:
@chrisrohr 
Thanks for creating the PR, and the fact that you included a test for it is perfect.  It's a good practice having tests created for all bug fixes.

I have just one change request.  The practice in this library is to simply call ```urlencode(String)``` directly, no reason to call ```getTagForPath(String)``` that simply calls ```urlencode(String)```, could you make this change?

chrisrohr:
@gmessner Yep I sure can.  I'll have the fix up in a few minutes.

chrisrohr:
@gmessner Update has been made and pushed.  Thank you

gmessner:
@chrisrohr 
I've merged your PR, I'll get a release out the next day or so.

Issue #382, opened 2019-06-10
Null pointer exception on export download


PhilippeVienne:
With gitlab.com the "Content-Disposition" is not given, it causes a null pointer here on line 112:
https://github.com/gitlab4j/gitlab4j-api/blob/master/src/main/java/org/gitlab4j/api/ImportExportApi.java#L111-L113

gmessner:
@PhilippeVienne 
This has been released in 4.11.8.  Please verify and close when you have the time.

Thanks again for your contribution.

PhilippeVienne:
@gmessner 
Thanks, it works :smile:

PR #383, opened 2019-06-10
Fix for null pointer #382

PhilippeVienne:


gmessner:
@PhilippeVienne 
Thanks for the PR, weird that GitLab excludes the ```Content-Disposition``` header.  That being said, I would prefer that a parameter is added for the filename, this would allow the user to control the name of the file (we already give them an option of setting the directory).
```
public File downloadExport(Object projectIdOrPath, File directory, String filename)
```
The current method ```downloadExport(Object projectIdOrPath, File directory)``` would simply call:
```
return (downloadExport(projectIdOrPath, directory, null);
```

In ```downloadExport(Object projectIdOrPath, File directory, String filename)``` if both the user provided filename and the ```Content-Disposition``` header are null, then a filename would need to be constructed. I would also like to make the constructed filename a bit more useful and patterned after the filename that is returned by non gitlab.com servers.  That filename looks like this:
```2019-06-08_03-46-178_gitlab4j_test-project_export.tar.gz```

My suggestion for constructing the filename is to use a format:

```projectIdOrPath``` is an Integer (project ID)::
```
String template = "%1$tY-%1$tm-%1$td_%1$tH-%1$tM-%1%tS_projectid-%2$d_export.tar.gz";
String filename = template.format(neew Date(), projectIdOrPath);
// filename = "2019-06-10_10-28-52_projectid_3115610_export.tar.gz"
```
```projectIdOrPath``` is a String instance, use it:
```
String template = "%1$tY-%1$tm-%1$td_%1$tH-%1$tM-%1$tS_%2$s_export.tar.gz";
String filename = String.format(template, new Date(), projectIdOrPath);
// filename = "2019-06-10_10-28-52_test-project_export.tar.gz"
```
```projectIdOrPath``` is a Project instance, use the project path:
```
String template = "%1$tY-%1$tm-%1$td_%1$tH-%1$tM-%1$tS_%2$s_export.tar.gz";
String filename = String.format(template, new Date(), testProject.getPath());
// filename = "2019-06-10_10-28-52_test-project_export.tar.gz"
```

**Comment by gmessner** [comment]
It might be better to use getPathWithNamespace() for the project instance and make sure to replace any slashes ```/``` with an underscore ```_```

**Comment by PhilippeVienne** [comment]
Working on requested changes :smiley:

File: src/main/java/org/gitlab4j/api/ImportExportApi.java [hunk 1]
```diff
@@ -108,18 +125,32 @@ public File downloadExport(Object projectIdOrPath, File directory) throws GitLab
...
+                    } else if(projectIdOrPath instanceof String) {
+                        String template = "%1$tY-%1$tm-%1$td_%1$tH-%1$tM-%1$tS_%2$s_export.tar.gz";
+                        filename = String.format(template, new Date(), projectIdOrPath);
+                        // filename = "2019-06-10_10-28-52_test-project_export.tar.gz"
+                    } else if(projectIdOrPath instanceof Integer) {
+                        String template = "%1$tY-%1$tm-%1$td_%1$tH-%1$tM-%1%tS_projectid-%2$d_export.tar.gz";
+                        filename = String.format(template, new Date(), projectIdOrPath);
+                        // filename = "2019-06-10_10-28-52_projectid_3115610_export.tar.gz"
+                    } else {
+                        // Fallback for strange types
```
gmessner: This will never occur as the call to ```getProjectIdOrPath(projectIdOrPath)``` above will throw an exception if it is not a valid Project instance,  project path, or project ID (Integer).

gmessner:
@PhilippeVienne 
Once again, thanks for the contribution.  Just one small request above, everything else looks great.

Also, would be good to add integration tests for these other conditions.

Issue #387, opened 2019-06-14
Support masked attribute for CI Variables


PhilippeVienne:
Support the `masked` attribute for CI Variables as described in https://docs.gitlab.com/ee/api/project_level_variables.html#create-variable

Working on PR

gmessner:
@PhilippeVienne 
This has been released in 4.11.9.  In the process of the release, I found another problem with the GitLab documentation, the ```masked``` field is never returned in the JSON from the server, so will always be null.  I have changed the integration tests accordingly.  

Please verify and close when you have time.

gmessner:
@PhilippeVienne 
Just wondering if you could verify and close this issue?

gmessner:
@PhilippeVienne 
Since the fix has been released I am closing this issue.  If you find any problems feel free to re-open.

PR #388, opened 2019-06-14
Add support for  CI masked variables

PhilippeVienne:
Refer to issue #387

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/Variable.java
+++ b/src/main/java/org/gitlab4j/api/models/Variable.java
@@ -14,6 +14,8 @@ public class Variable {
     private String value;
     @JsonProperty("protected")
     private Boolean isProtected;
+    @JsonProperty("masked")
+    private Boolean isMasked;
     private String environmentScope;
 
     public Variable() {
@@ -46,6 +48,14 @@ public class Variable {
 
     public void setProtected(Boolean isProtected) {
         this.isProtected = isProtected;
+    }
+
+    public Boolean getMasked() {
+        return isMasked;
+    }
+
+    public void setMasked(Boolean masked) {
+        isMasked = masked;
     }
 
     public String getEnvironmentScope() {
--- a/src/test/java/org/gitlab4j/api/TestProjectApi.java
+++ b/src/test/java/org/gitlab4j/api/TestProjectApi.java
@@ -677,7 +677,6 @@ public class TestProjectApi extends AbstractIntegrationTest {
         assertNotNull(variable);
         assertEquals(key, variable.getKey());
         assertEquals(value, variable.getValue());
-        assertFalse(variable.getMasked());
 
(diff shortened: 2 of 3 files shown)


gmessner:
@PhilippeVienne 
Thanks for your contribution.  

Just one small request, I noticed there is not a bean test for the variable details. Could you add the following file to src/test/resources/org/gitlab4j/api:

project-variable-details.json
```
{
    "key": "TEST_VARIABLE_1",
    "variable_type": "env_var",
    "value": "TEST_1",
    "protected": false,
    "masked": true
}
```
And then the following test to ```TestGitLabApiBeans.java```:
```
    @Test
    public void testProjectVariableDetails() throws Exception {
        Variable variable = unmarshalResource(Variable.class, "project-variable-details.json");
        assertTrue(compareJson(variable, "project-variable-details.json"));
    }
```

src/main/java/org/gitlab4j/api/models/Variable.java
```diff
@@ -14,6 +14,8 @@
     private String value;
     @JsonProperty("protected")
     private Boolean isProtected;
+    @JsonProperty("masked")
+    private Boolean isMasked;
```
gmessner: @PhilippeVienne 
No need to name this ```isMasked```, just name it ```masked``` and remove the ```@JsonProperty("masked")``` line.

The reason ```isProtected``` was used is because ```protected``` is a reserved keyword in Java.

src/test/java/org/gitlab4j/api/TestProjectApi.java
```diff
@@ -672,11 +672,12 @@ public void testVariables() throws GitLabApiException {
 
         String key = TEST_VARIABLE_KEY_PREFIX + HelperUtils.getRandomInt() + "_" +  HelperUtils.getRandomInt();
         String value = "TEST_VARIABLE_VALUE_" + HelperUtils.getRandomInt() + "_" +  HelperUtils.getRandomInt();
-        Variable variable = gitLabApi.getProjectApi().createVariable(testProject, key, value, null, null);
+        Variable variable = gitLabApi.getProjectApi().createVariable(testProject, key, value, null, null, null);
 
         assertNotNull(variable);
         assertEquals(key, variable.getKey());
         assertEquals(value, variable.getValue());
+        assertFalse(variable.getMasked());
```
gmessner: @PhilippeVienne 
Not certain on this but I believe ```variable.getMasked()``` will return ```null``` in this case.

PhilippeVienne: Yes correct and that's why a second commit remove it

gmessner:
@PhilippeVienne 
Getting ready to do a release, I'm going to go ahead and merge this and then add the test I mention in my first commit.  This will be included in the upcoming release.

PR #439, opened 2019-09-26
Adding withLastActivityAt(Date lastActivityAt) to Project

stefano-salmaso:


gmessner:
@stefano-salmaso 
Why is ```withLastActivityAt(Date lastActivityAt)``` needed?  The ```withXXXX()``` methods in ```Project``` are used when a project is created or updated and it is not a value the user can set, it is a value that is set and persisted server side only.

stefano-salmaso:
HI @gmessner 
I've added this method because, in a unit test, I created a project and it would be useful to have it.
In any case, I understood your point.
Just a question, the `setLastActivityAt (Date lastActivityAt) method is public so it could still be changed, am I wrong?

Feel free to close / reject this MR.

Thanks
Stefano

gmessner:
@stefano-salmaso 
Yes you can use ```setLastActivityDayAt(Date lastActivityAt)```.

Thank you for the contribution, but I will be closing the PR without merging.

PR #451, opened 2019-10-16
Update archive_builds_in_human_readable to String instead of boolean

jabby:
Hello, 

According to the documentation of gitlab, it seems that `archive_builds_human_readable` is a String and not a boolean. 
https://docs.gitlab.com/ee/api/settings.html#list-of-settings-that-can-be-accessed-via-api-calls

This PR fix the enum configuration for `archive_builds_in_human_readable`


gmessner:
@jabby 
Thanks for your contribution.  I will merge, but a release will not happen for a few days, need more significant changes to do a release.  WIll let you know when the release is done.

jabby:
@gmessner I understand. No problem I will use a local snapshot for my development until the release.

gmessner:
@jabby 
Release 4.12.14 is now available with this change included.

PR #453, opened 2019-10-21
Create commit on existing branch

oraclewalid:


gmessner:
@oraclewalid 
Just wondering why you can't just call the following with ```startBranch``` set to null?  It will do the exact same thing as the new methods you implemented.
```
public Commit createCommit(Object projectIdOrPath, String branch, String commitMessage, String startBranch, String authorEmail, String authorName, CommitAction action)

public Commit createCommit(Object projectIdOrPath, String branch, String commitMessage, String startBranch,  String authorEmail, String authorName, List<CommitAction> actions)
```

oraclewalid:
@gmessner 
- First reason:, I spend many time to find how to do that, and I found response in unit tests,  so with these functions will make it more intuitive.
- Second reason: Using null make code more difficult to understand and ugly.
Thanks

gmessner:
@oraclewalid 
In response to the use of null making it more difficult to understand and ugly, the same is true for the ```authorEmail``` and ```authorName``` parameters, those can also be null.  We cannot create methods simply because we do not like passing null, this library would quickly become way too large and more difficult to understand and maintain because of the multiple methods for similar functionality.

I would prefer to simply clarify this in the Javadoc documentation for the existing ```createCommit()``` methods rather than have multiple methods, this will help make it more understandable and easier to maintain.

gmessner:
@oraclewalid 
I have released a new version of GitLab4J-API (4.12.17)  with a simplified way to create commits.  Take a look at ```TestCommitsApi.testCreateCommitWithPayload()``` it will show you how to use this new approach for creating commits.

Thank you for your contribution, while it will not be merged it is what drove me to develop a simpler way to create commits.

oraclewalid:
Thanks for this elegant solution , I will  upgrade the version to use this function

PR #456, opened 2019-10-25
docs: fix the gradle version artifact

ojacquemart:
The actual docs refers to an inexisting release version


gmessner:
@ojacquemart 
Thanks for noticing this typo and fixing it.

ojacquemart:
Thanks for this great project. It is so cool.

PR #460, opened 2019-10-28
Update property name from approvals missing to approvals left

jabby:
According to the latest documentation, the approvals missing property does not exist for merge request. The name I found in the documnetation is approvals left. 

See [documentation here](https://docs.gitlab.com/ee/api/merge_request_approvals.html#get-configuration-1)

This PR change the property name from `approvals missing` to `approvals left`. I build locally a snapshot and try it on a private gitlab with starter license. It works like a charm. 

I suspect that oldest version of API use the term `approvals missing` but I didn't find any clue about my thougth.

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/MergeRequest.java
+++ b/src/main/java/org/gitlab4j/api/models/MergeRequest.java
@@ -55,7 +55,7 @@ public class MergeRequest {
 
     // The approval fields will only be available when listing approvals, approving  or unapproving a merge reuest.
     private Integer approvalsRequired;
-    private Integer approvalsMissing;
+    private Integer approvalsLeft;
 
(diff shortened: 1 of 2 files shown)


gmessner:
@jabby 
First off, thank you for your contribution.

I did a bit of research to be sure that the switch from ```approvals_missing``` to ```approvals_left``` was not too recent.  It was changed way back in June 2016, so this is OK for me to merge.  If it was recent, I would have needed to support both and indicate that ```approvals_missing``` is deprecated.

Good find.

jabby:
You're welcome. I use this project and sometimes it does not work as expected. I will try to provide patches when I can. 

Just one question, how did you find the information about June 2016. I think my search was not enough good.  My goal is to provide more information for future PRs.

PR #462, opened 2019-10-31
Update settings to latest version of gitlab api.

jabby:
According to the documentation there is many new settings available.
This commit update all settings and documentation to the latest version.

It's actually a work in progress as there is few TODOs in the code.

Basically the first verification to do is to validate that this modification is retro-compatible
with olders version of library.

I will update this PR next day. Feel free to comment my work in progress.

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/Setting.java
+++ b/src/main/java/org/gitlab4j/api/models/Setting.java
@@ -19,40 +19,81 @@ public enum Setting {
      */
     ADMIN_NOTIFICATION_EMAIL(String.class),
 
-    /** Where to redirect users after logout. */
+    /**
+     * Where to redirect users after logout.
+     */
     AFTER_SIGN_OUT_PATH(String.class),
 
-    /** Text shown to the user after signing up */
+    /**
+     * Text shown to the user after signing up
+     */
     AFTER_SIGN_UP_TEXT(String.class),
 
-    /** required by: akismet_enabled API key for akismet spam protection. */
+    /**
+     * required by: {@link #AKISMET_ENABLED} API key for Akismet spam protection.
+     */
     AKISMET_API_KEY(String.class),
 
     /**
-     * (If enabled, requires: akismet_api_key) Enable or disable akismet spam
+     * (<strong>If enabled, requires:</strong> {@link #AKISMET_API_KEY}) Enable or disable Akismet spam
      * protection.
      */
     AKISMET_ENABLED(Boolean.class),
 
-    /** (Premium) Set to true to allow group owners to manage LDAP */
+    /**
+     * (PREMIUM | SILVER) Set to true to allow group owners to manage LDAP
+     */
     ALLOW_GROUP_OWNERS_TO_MANAGE_LDAP(Boolean.class),
 
-    /** Allow requests to the local network from hooks and services. */
+    /**
+     * Allow requests to the local network from hooks and services.
+     * @deprecated Use allow_local_requests_from_web_hooks_and_services instead
+     */
+    @Deprecated
     ALLOW_LOCAL_REQUESTS_FROM_HOOKS_AND_SERVICES(Boolean.class),
 
-	/**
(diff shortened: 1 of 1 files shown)


src/main/java/org/gitlab4j/api/models/Setting.java
```diff
@@ -67,54 +108,60 @@
...
-    CLIENTSIDE_SENTRY_DSN(String.class),
-
+    COMMIT_EMAIL_HOSTNAME(String.class),
+    
     /**
-     * (If enabled, requires: clientside_sentry_dsn) Enable Sentry error reporting
-     * for the client side.
+     * Container Registry token duration in minutes.
      */
-    CLIENTSIDE_SENTRY_ENABLED(Boolean.class),
```
gmessner: This should be deprecated not removed.

src/main/java/org/gitlab4j/api/models/Setting.java
```diff
@@ -67,54 +108,60 @@
...
+     * available to projects if the project namespace’s plan includes the feature
+     * or if the project is public.
      */
     CHECK_NAMESPACE_PLAN(Boolean.class),
 
     /**
-     * required by: clientside_sentry_enabled Clientside Sentry Data Source Name.
+     * Custom hostname (for private commit emails).
      */
-    CLIENTSIDE_SENTRY_DSN(String.class),
```
gmessner: This should be deprecated not removed.

src/main/java/org/gitlab4j/api/models/Setting.java
```diff
@@ -217,485 +307,629 @@
...
     RESTRICTED_VISIBILITY_LEVELS(String[].class),
 
     /**
-     * The minimum allowed bit length of an uploaded RSA key. Default is 0 (no
-     * restriction). -1 disables RSA keys.
+     * The minimum allowed bit length of an uploaded RSA key. Default is 0 (no restriction).
+     * -1 disables RSA keys.
      */
-    RSA_KEY_RESTRICTION(Integer.class),
+    RSA_KEY_RESTRICTION(Boolean.class),
```
gmessner: The documentation indicates that ```0``` or ```-1``` is returned, why change this to ```Boolean.class````

src/main/java/org/gitlab4j/api/models/Setting.java
jabby: I updated my patch. It was a mistake.

src/main/java/org/gitlab4j/api/models/Setting.java
jabby: I updated my patch. It was a mistake.

src/main/java/org/gitlab4j/api/models/Setting.java
jabby: Good catch. My bad.

jabby:
@gmessner I updated the change accordingly to your comments. I also updated few more things and dig a little for fields with no documentation. I updated javadoc when I found some useful informations. 

There is one task to do before merging this patch. See https://github.com/gitlab4j/gitlab4j-api/pull/462/files#diff-999bbdddeac9c1f965364d46f7c74cdeR88
I don't know which type we should use here.

gmessner:
@jabby 
I have a way to have it support both ```String``` and ```String[]```, but it would be more work for me to explain how to do it rather than do it myself.

My suggestion would be for you to finish everything else and let me know it is ready to be merged.  Once merged I will add support for the dual type and keep you updated on how it was accomplished.

jabby:
For my point of view, except about the support of `String` and `String[]`, this patch can be merged.
So feel free to merged and provide the additional change. I will check your modification and learn how you manage this case.

gmessner:
@jabby 
I am currently traveling, will be back late 11/6/2019 will merge then and add support for the ```String/String[]``` dual type.

jabby:
Ok no problem. Have a nice travel.

PR #466, opened 2019-11-05
Add source attribute to PipelineEvent.ObjectAttributes

maguro:


gmessner:
@maguro 
Thanks for your contribution.  Should be doing a release in a day or two.

PR #469, opened 2019-11-08
Add commit last pipeline

Gaff:


src/main/java/org/gitlab4j/api/models/Commit.java
```diff
@@ -162,6 +163,9 @@ public void setUrl(String url) {
         this.url = url;
     }
 
+    public Pipeline getLastPipeline() { return lastPipeline; }
+
+    public void setLastPipeline(Pipeline lastPipeline) { this.lastPipeline = lastPipeline; }
```
gmessner: @Gaff 
To be consistent with the style used in this library change these to multiline methods:
```
public Pipeline getLastPipeline() {
    return lastPipeline;
}

public void setLastPipeline(Pipeline lastPipeline) {
    this.lastPipeline = lastPipeline;
}
```

gmessner:
@Gaff 
Because I'd like to include this in the release that will b done later today I am going to go ahead and merge this.  I will resolve any conflicts.

Thanks again for the contribution.

Gaff:
Sorry - had logged off for the weekend so didn't pick this up. Thanks for merging!

PR #491, opened 2019-12-31
BUG FIX in accept mr API.

home3k:
GitLab V3 API using 'merge_request' path arg.


gmessner:
@home3k 
Looking at the v3 documentation it says that it should be "merge_requests", see https://gitlab.com/gitlab-org/gitlab-foss/blob/8-16-stable/doc/api/merge_requests.md#accept-mr

What version of GitLab server are you using?  What I don't want to do is break this for others that are using the v3 API.

BTW, This library will remove support for GitLab API v3 sometime soon as support for it was removed from GitLab almost 2 years ago and v3 was replaced by v4 almost 3 years ago.

gmessner:
@home3k Have not heard back from you so I am closing this merge request.

Issue #486, opened 2019-12-18
Support "Add note to existing merge request thread" and "Modify an existing merge request thread note" in Discussions API


derekyyyan:
Hi,

First off, I'd like to thank you for creating and consistently maintaining this library.

I'd like to use these two GitLab Discussions API endpoints that this library currently doesn't support:
1. [Add note to existing merge request thread](https://docs.gitlab.com/ee/api/discussions.html#add-note-to-existing-merge-request-thread)
2. [Modify an existing merge request thread note](https://docs.gitlab.com/ee/api/discussions.html#modify-an-existing-merge-request-thread-note)

I'd be glad to implement these endpoints myself. What is the preferred process for contributing to this library? Anything else I need to know?

Thanks!

gmessner:
@derekyyyan 
Thanks for offering to implement, but I had some time today and have implemented support for the merge request thread note methods.  This is available in release 4.14.7.  Here are the new methods:
```
DiscuttionasApi.addMergeRequestThreadNote(Object projectIdOrPath,
  Integer mergeRequestIid, String discussionId, String body, Date createdAt)

DiscussionsApi.modifyMergeRequestThreadNote(Object projectIdOrPath,
  Integer mergeRequestIid, String discussionId, Integer noteId, String body, Boolean resolved)

DiscussionsApi.deleteMergeRequestThreadNote(Object projectIdOrPath,
  Integer mergeRequestIid, String discussionId, Integer noteId)
```

Please verify and close this issue when you have time. 

I need to add the contributing, issue templates, and pull request template pages to document how to contribute, I hope to get to these in the near future.  Once I have those done it will be easier for others to contribute.

derekyyyan:
Thank you so much @gmessner !
I'm on vacation right now so I won't be able to verify the changes until next week, but I'll do that and close the issue once I get back.

derekyyyan:
The new endpoints weren't working, and then I found some typos. I made a pull request fixing those typos: https://github.com/gitlab4j/gitlab4j-api/pull/492

gmessner:
@derekyyyan 
Thanks for verifying, I have merged your PR and will be doing a release in the next day or so.  I will let you know when it is available.

gmessner:
@derekyyyan 
Release 4.14.10 is now available with your fix.  Please re-verify and close when you have time.

derekyyyan:
Thanks! I'll be able to verify later today.

derekyyyan:
[...]

PR #492, opened 2020-01-03
Fix typo (#486)

derekyyyan:


gmessner:
Thanks for the fix, will merge and get a release out in the next day or two.

PR #499, opened 2020-01-18
Add parameter with_content=1

marcelomrwin:
If this parameter is not passed the content of wiki will not return and content will be null


gmessner:
@marcelomrwin 
Thanks for your contribution.  However, there is a typo so the code will not compile (two commas).  Additionally, the with_content should have been a parameter to the method, I have updated the code to include methods that take withContent and have also added Pager and Stream methods to bring it up to date with this project's current conventions.  This code is available and a release will be done in a few days.

gmessner:
@marcelomrwin 
Release 4.14.12 is available and has added methods to the ```WikisApi``` that take the ```withContent``` parameter.

marcelomrwin:
Thanks

Em dom, 19 de jan de 2020 às 03:19, Greg Messner <notifications@github.com>
escreveu:

PR #504, opened 2020-02-05
Add support for multiple assignees in Merge Requests

eedzjee:
GitLab supports multiple assignees for Merge Requests.
I added this field to the MergeRequest object, so it can be used.

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/MergeRequest.java
+++ b/src/main/java/org/gitlab4j/api/models/MergeRequest.java
@@ -14,6 +14,7 @@ public class MergeRequest {
     private Boolean allowMaintainerToPush;
     private Integer approvalsBeforeMerge;
     private Assignee assignee;
+    private List<Assignee> assignees;
     private Author author;
     private List<Diff> changes;
     private Date closedAt;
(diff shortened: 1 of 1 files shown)


src/main/java/org/gitlab4j/api/models/MergeRequest.java
```diff
@@ -86,6 +87,14 @@ public void setApprovalsBeforeMerge(Integer approvalsBeforeMerge) {
...
     }
 
+    public List<Assignee> getAssignees() {
+        return assignees;
+    }
+
+    public void setAssignees(List<Assignee> assignees) {
+        this.assignees = assignees;
+    }
+
```
gmessner: @eedzjee These should be moved to after ```setAssignee(Assignee assignee)```

eedzjee: You're right! I've updated the file.

gmessner: Thanks for making the change.   The next release is scheduled for 2/10/2020, will let you know when it is released.

Issue #505, opened 2020-02-07
support returning Pager for getCommitRefs


blacksteed232:
Hi, we just bumped into a use case (weird one) that the number of `commit refs` returned is greater than the default page size (96) of gitlab4j so I thought it would make sense to have a method that returns a Pager of `commit refs` rather than a list. That would be great!

UPDATE: Attempted to do this with a pull request #506

gmessner:
@blacksteed232 I have merged your pull request and added additional ```getCommitRefs()```.  The ```getCommitRefs()``` methods that return ```List<CommitRef>``` now return a List of **all** CommitRefs regardless of the count.

GitLab4J-API Release 4.14.16 is now available, please verify and close this issue when you have time.

blacksteed232:
@gmessner do you know how long it usually takes to appear on maven central?

gmessner:
It can take up to 24 hours depending on activity.

gmessner:
@blacksteed232 I just checked and 4.14.16 is available, it is just not indexed yet.  I believe if you specify 4.14.16 in your pom or build.gradle file it will find it.

blacksteed232:
@gmessner looks good to me. Closing this one now. Thank you!

PR #506, opened 2020-02-07
add a method to get pager of commit refs

blacksteed232:
@gmessner Regarding #505 . This is my novice attempt (without integration test since I don't quite know how you like to setup test for Pager -- create multiple branches from a commit?). Hope this helps!


gmessner:
@blacksteed232 
Thanks for your contribution.  I will merge and make some adjustments as we now follow a pattern where we also include a ```getCommitRefsStream()``` method and the method that returns a List calls the method that returns a Pager and uses the ```al()``` method of Pager to return all objects in a List.

Will let you know when the release is available.

PR #511, opened 2020-02-17
Add method for getting MR with diverged_commits_count attribute

### LINKED ISSUE #510: Missing diverged_commits_count attribute on merge requests
It looks like Merge Requests are missing the optional attribute `diverged_commits_count` that is only returned when specifying `include_diverged_commits_count=true` when [getting a single MR](https://docs.gitlab.com/ee/api/merge_requests.html#get-single-mr).

This is similar to the `rebase_in_progress` attribute.

krische:
This adds a new method `getDivergedCommitsCount()` to the MergeRequestApi class that gets a single MR and includes the `include_diverged_commits_count` query parameter. This is similar to the behavior of `getRebaseStatus()`.

I wasn't sure how you want to test this, so looking for feedback on that.

Fixes #510

### DIFF
--- a/src/main/java/org/gitlab4j/api/MergeRequestApi.java
+++ b/src/main/java/org/gitlab4j/api/MergeRequestApi.java
@@ -250,7 +250,32 @@ public class MergeRequestApi extends AbstractApi {
      * @throws GitLabApiException if any exception occurs
      */
     public MergeRequest getMergeRequest(Object projectIdOrPath, Integer mergeRequestIid) throws GitLabApiException {
-        Response response = get(Response.Status.OK, null, "projects", getProjectIdOrPath(projectIdOrPath), "merge_requests", mergeRequestIid);
+        return getMergeRequest(projectIdOrPath, mergeRequestIid, null, null, null);
+    }
+
+    /**
+     * Get information about a single merge request.
+     *
+     * <pre><code>GitLab Endpoint: GET /projects/:id/merge_requests/:merge_request_iid</code></pre>
+     *
+     * @param projectIdOrPath the project in the form of an Integer(ID), String(path), or Project instance
+     * @param mergeRequestIid the internal ID of the merge request
+     * @param renderHtml if true response includes rendered HTML for title and description, can be null
+     * @param includeDivergedCommitCount if true response includes the commits behind the target branch, can be null
+     * @param includeRebaseInProgress if true response includes whether a rebase operation is in progress, can be null
+     * @return a MergeRequest instance as specified by the parameters
+     * @throws GitLabApiException if any exception occurs
+     */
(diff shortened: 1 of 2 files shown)


krische:
So I actually changed this up a little bit. Instead of adding yet another method for these optional parameters when getting a single MR (`getMergeRequest()`); I added an extra varargs parameter of the available options.

This way a new method isn't needed in `MergeRequestApi` every time GitLab adds another query parameter to that endpoint. And it also means users can make one request with several of those parameters specified.

gmessner:
@krische Thanks again for your contribution.

However we don't use this varargs pattern anywhere in this library, The pattern we use is if the parameter count is <= 5 we create a method to take those 5 parameters. and then you simply pass ```null``` as the parameter value if you don't want it set.  If the parameter count > 5 we create a filter class that uses ```withXXX()``` methods to set the parameters. So, in this case,, it would look something like this:

```
/**
 * Get information about a single merge request.
 *
 * <pre><code>GitLab Endpoint: GET /projects/:id/merge_requests/:merge_request_iid</code></pre>
 *
 * @param projectIdOrPath the project in the form of an Integer(ID), String(path), or Project instance
 * @param mergeRequestIid the internal ID of the merge request
 * @param renderHtml if true response includes rendered HTML for title and description, can be null
 * @param includeDivergedCommitCount if true response includes the commits behind the target branch, can be null
 * @param includeRebaseInProgress if true response includes whether a rebase operation is in progress, can be null
 * @return a MergeRequest instance as specified by the parameters
 */
public MergeRequest getRebaseStatus(Object projectIdOrPath, Integer mergeRequestIid,
       Boolean, renderHtml, Boolean includeDivergedCommitCount , Boolean includeRebaseInProgress  ) throws GitLabApiException {

    GitLabApiForm queryParams = new GitLabApiForm()
        .withParam("render_html", renderHtml)
        .withParam("include_diverged_commits_count", includeDivergedCommitCount)
        .withParam("include_rebase_in_progress", includeRebaseInProgress);

    Response response = get(Response.Status.OK, queryParams.asMap(),
	  projects", getProjectIdOrPath(projectIdOrPath), "merge_requests", mergeRequestIid);
    return (response.readEntity(MergeRequest.class));
}
```

The other convenience methods would simply call this method passing null for the parameters they don't want to set.  We also try to follow the order of the parameters that are in the GitLab API docs.

krische:
@gmessner I applied your recommended changes. I assumed you wanted an additional overloaded method of `getMergeRequest()` so as not to break signature of the existing method.

gmessner:
@krische Good catch, I should have included in my comment that we also include a no parameter method.  This looks good, I'll merge now, but a release will not be done until 2/22/2020.  If you have a pressing need for this functionality let me know and I'll get a release out sooner.

PR #515, opened 2020-02-20
Add search branch

hrotkogabor:
Since Gitlab 10.6 the Branches API has a search param for branch search.
See: https://gitlab.com/gitlab-org/gitlab-foss/-/merge_requests/17005/diffs
Documentation: https://docs.gitlab.com/ee/api/branches.html
Example call: http://localhost:8090/api/v4/projects/1/repository/branches?search=test

### DIFF
--- a/README.md
+++ b/README.md
@@ -500,7 +500,11 @@ List<Release> releases = gitLabApi.getReleasesApi().getReleases(projectId);
 #### RepositoryApi
 ```java
 // Get a list of repository branches from a project, sorted by name alphabetically
-List<Branch> branches = gitLabApi.getRepositoryApi().getBranches();
+List<Branch> branches = gitLabApi.getRepositoryApi().getBranches(projectId);
+```
+```java
+// Search repository branches from a project, by name
+List<Branch> branches = gitLabApi.getRepositoryApi().searchBranch(projectId, searchTerm);
 ```
 
 #### RepositoryFileApi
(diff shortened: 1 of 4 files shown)


gmessner:
@hrotkogabor Thank you for your contribution.  This will be useful to add but it needs to follow our standard of a 1:1 relationship with the GitLab API and our standard pattern of having a set of 3 methods when getting lists of items.  For this the methods should be as follows:
```
public List<Branch> getBranches(Object projectIdOrPath, String search)
public Pager<Branch> getBranches(Object projectIdOrPath, String search, int itemsPerPage)
public Stream<Branch> getBranchesStream(Object projectIdOrPath, String search)
```

The method that returns ```List<Branch>``` would simply call the method that returns ```Pager<Branch>``` as follows:
```
return (getBranches(projectIdOrPath, search, getDefaultPerPage()).all());
```
And the method that returns ```Stream<Branch>``` would simply call the method that returns ```Pager<Branch>``` as follows:
```
return (getBranches(projectIdOrPath, search, getDefaultPerPage()).stream());
```

The ```getBranches``` methods that don't take a search parameter could be changed to call the new methods, passing null as the search param.

hrotkogabor:
I made the refactor. Please check.

gmessner:
@hrotkogabor Thanks for the updates, I've merged and will do a release in the next day or so.

gmessner:
@hrotkogabor Release 4.14.18 is now available.

PR #520, opened 2020-03-03
Implement AutoCloseable to close the underlying client.

krische:
I'm not certain if this is completely necessary, but it seems like a "doesn't hurt" sort of thing.

I did notice that the the [Client interface docs](https://docs.oracle.com/javaee/7/api/javax/ws/rs/client/Client.html) say the client should be closed before being disposed; so I made the `GitLabApi` and subsequent `GitLabApiClient` classes implement `AutoCloseable` to close the `JerseyClient` that is used.

### DIFF
--- a/src/main/java/org/gitlab4j/api/GitLabApi.java
+++ b/src/main/java/org/gitlab4j/api/GitLabApi.java
@@ -1,5 +1,6 @@
 package org.gitlab4j.api;
 
+import java.io.Closeable;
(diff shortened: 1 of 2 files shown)


gmessner:
@krishe  I also believe this "doesn't hurt" and may take care of a potential memory leak, so am merging now.

gmessner:
@krische  GitLab4J-API release 4.14.19 is available with this functionality.

PR #524, opened 2020-03-04
Fix serialization error with new params.

krische:
It looks like I forgot to add `.asMap()` to the new merge request acceptance parameters. Without this, it blows up with: `org.gitlab4j.api.GitLabApiException: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: No serializer found for class org.gitlab4j.api.GitLabApiForm and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS)`

This change fixes that.


gmessner:
@krische Good find.  As you have discovered Form is used for constructing query params (this issue) which are expected to be a ```MultiValueMap```, if you don't call ```asMap()``` the object is considered a payload for the request.  Merging now.

gmessner:
@krische  GitLab4J-API release 4.14.19 is available with this fix.

PR #525, opened 2020-03-06
fix passing author_id param in MergeRequestFilter

rnemykin:


src/main/java/org/gitlab4j/api/models/MergeRequestFilter.java
```diff
@@ -11,7 +9,11 @@
 import org.gitlab4j.api.Constants.SortOrder;
 import org.gitlab4j.api.GitLabApiForm;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
+import java.util.Date;
+import java.util.List;
+
```
gmessner: @rnemykin Would prefer to not reorder the imports in this fashion, this library uses the Eclipse ordering pattern, which was how they were previously ordered.  The Eclipse ordering puts the Java imports first followed by the imports that belong to the library, the remaining imports are alphabetically ordered and then followed by the static imports.

rnemykin: Yeah, missed that, it was IDEA optimize imports options during pushing.
fixed with Eclipse convention

gmessner:
@rnemykin Thanks for the update.. I will be merging now.  I am waiting on another pull request, as soon as that is complete I will be doing a release, will let you know when that occurs.

gmessner:
@rnemykin  GitLab4J-API release 4.14.19 is available with this functionality.

PR #526, opened 2020-03-06
Add createIssueDiscussion endpoint

svendroid:
Allows to create a new comment for an issue by using the DiscussionsApi:
https://docs.gitlab.com/ee/api/discussions.html#create-new-issue-thread

### DIFF
--- a/src/main/java/org/gitlab4j/api/DiscussionsApi.java
+++ b/src/main/java/org/gitlab4j/api/DiscussionsApi.java
@@ -671,4 +671,26 @@ public class DiscussionsApi extends AbstractApi {
         delete(Response.Status.OK, null, "projects", getProjectIdOrPath(projectIdOrPath),
                 "merge_requests", mergeRequestIid, "discussions", discussionId, "notes", noteId);
     }
+
+    /**
+     * Creates a new thread to a single project issue. This is similar to creating a note but other comments (replies) can be added to it later.
+     *
+     * <pre><code>GitLab Endpoint: POST /projects/:id/issues/:issue_iid/discussions</code></pre>
+     *
+     * @param projectIdOrPath projectIdOrPath the project in the form of an Integer(ID), String(path), or Project instance
+     * @param issueIid The IID of an issue
+     * @param body the content of the discussion
+     * @param createdAt (optional) date the discussion was created (requires admin or project/group owner rights)
+     * @return a Discussion instance containing the newly created discussion
+     * @throws GitLabApiException if any exception occurs during execution
+     */
+    public Discussion createIssueDiscussion(Object projectIdOrPath, Integer issueIid, String body, Date createdAt) throws GitLabApiException {
+        GitLabApiForm formData = new GitLabApiForm()
+                .withParam("body", body, true)
+                .withParam("created_at", createdAt);
+        Response response = post(Response.Status.CREATED, formData,
+                "projects", getProjectIdOrPath(projectIdOrPath), "issues", issueIid, "discussions");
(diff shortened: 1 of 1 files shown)


gmessner:
@svendroid Thanks for the contribution.  To make this complete the following methods should also be added:
```
public Note addIssueThreadNote(Object projectIdOrPath, Integer issueIid,
	    String discussionId, String body, Date createdAt) 
 public Note modifyIssueThreadNote(Object projectIdOrPath, Integer issueIid,
	    String discussionId, Integer noteId, String body)
public void deleteIssueThreadNote(Object projectIdOrPath, Integer issueIid,
	    String discussionId, Integer noteId)
```
The documentation at https://docs.gitlab.com/ee/api/discussions.html#add-note-to-existing-snippet-thread for "Add note to existing issue thread" indicates that a ```note_id``` should be included but I beleive this is incorrect as the call itself creates the note.

svendroid:
I added methods to add, modify & delete notes for a discussion of an issue. The gitlab API also use the term thread for a discussion. I went with Discussion because the rest Endpoint is called discussions and the model inside gitlab4j is also a Discussion.

> The documentation at https://docs.gitlab.com/ee/api/discussions.html#add-note-to-existing-snippet-thread for "Add note to existing issue thread" indicates that a note_id should be included but I beleive this is incorrect as the call itself creates the note.
The link describes the notes for a snippet discussion. Maybe it is a copy paste error? Did you mean https://docs.gitlab.com/ee/api/discussions.html#add-note-to-existing-issue-thread? This describes the notes regarding notes on issue discussions.

gmessner:
@svendroid Yes, I referenced the wrong link, and thanks for adding the new methods.

That being said, please use the names as I provided as that is the pattern this library uses for each of the note types (commit and merge request), it is best to be consistent and more in line with the GitLab API docs.

svendroid:
@gmessner thanks for the feedback. I renamed the methods.

src/main/java/org/gitlab4j/api/DiscussionsApi.java
```diff
@@ -671,4 +671,86 @@ public void deleteMergeRequestThreadNote(Object projectIdOrPath, Integer mergeRe
...
+
+    /**
+     * Deletes an existing thread note of an issue.
+     *
+     * <pre><code>GitLab Endpoint: DELETE /projects/:id/issues/:issue_iid/discussions/:discussion_id/notes/:note_id</code></pre>
+     *
+     * @param projectIdOrPath projectIdOrPath the project in the form of an Integer(ID), String(path), or Project instance
+     * @param issueIid The IID of an issue
+     * @param discussionId the id of discussion
+     * @param noteId the id of the note
```
gmessner: Need to add the following to the Javadocs:
```
* @throws GitLabApiException if any exception occurs during execution
```

gmessner:
@svendroid Thanks for the quick turnaround.  Just one small change and it is set to go.  If you are not able to get to the change today, I will go ahead and merge and then make the change, because I want to get a release out today.

src/main/java/org/gitlab4j/api/DiscussionsApi.java
svendroid: added the javadoc for the exception

svendroid:
@gmessner fixed the javadoc. Should I squash the commits into one? or will this be done automatically when you merge?

gmessner:
@svendroid That was fast.  No need to squash, that will happen when I merge it.

gmessner:
@svendroid GitLab4J-API release 4.14.19 is available with this functionality.

PR #543, opened 2020-03-31
Add repositorysize and wikisize

marcojahn:
Project 
- add property `statistics.repository_size`
- add property `statistics.wiki_size`
- fix property for `statistics.lfs_objects_size` (plural "s" was missing)


gmessner:
@maguro Just one request, can you group the get and set method for each property together? Not sure how I missed this before but ```ProjectStatistics``` has all the get methods group separately from the set methods, which is not the convention this library uses.

marcojahn:
Resorted as requested.

gmessner:
@marcojahn Thanks for the quick turn-around.  Release 4.14.24 has been released with this functionality.

Issue #544, opened 2020-04-01
Search API on projects should allow ref parameter


ribardo0:
Hi, 

It is currently not possible in gitlab4j to search in a project on a specific branch.

We should have a method in SearchAPI.java
public List<?> projectSearch(Object projectIdOrPath, ProjectSearchScope scope, String search, String ref)

Documentation in Gitlab API doc : https://docs.gitlab.com/ee/api/search.html#project-search-api
See example using a specific branch in Gitlab API doc : https://docs.gitlab.com/ee/api/search.html#scope-blobs

Thanks in advance,

Olivier

gmessner:
@ribardo0 Release 4.14.25 is available with this PR, please verify and close when you have time.

ribardo0:
Works perfectly fine... unfortunately BranchesAPI (https://docs.gitlab.com/ee/api/branches.html) is not available in Gitlab4J, I will open another issue

PR #545, opened 2020-04-01
Adding ref parameter on SearchAPI on projects (#544)

ribardo0:
Hi,

I didn't had time to test on real gitlap instance (mvn package is OK).

Hope this helps

Olivier

### DIFF
--- a/src/main/java/org/gitlab4j/api/SearchApi.java
+++ b/src/main/java/org/gitlab4j/api/SearchApi.java
@@ -208,9 +208,29 @@ public class SearchApi extends AbstractApi {
      * @since GitLab 10.5
      */
     public List<?> projectSearch(Object projectIdOrPath, ProjectSearchScope scope, String search) throws GitLabApiException {
-        return (projectSearch(projectIdOrPath, scope, search, this.getDefaultPerPage()).all());
+        return (projectSearch(projectIdOrPath, scope, search, null, this.getDefaultPerPage()).all());
     }
 
+    /**
+     * Search within the specified project.  If a user is not a member of a project and the project is private,
+     * a request on that project will result to a 404 status code.
+     *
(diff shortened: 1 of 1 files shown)


src/main/java/org/gitlab4j/api/SearchApi.java
```diff
@@ -226,29 +246,58 @@ public SearchApi(GitLabApi gitLabApi) {
...
+     * @param search the search query
+     * @param ref the name of a repository branch or tag to search on. The project’s default branch is used by 
+     *             default. This is only applicable for scopes: commits, blobs, and wiki_blobs.
      * @param itemsPerPage the number of items that will be fetched per page
      * @return a Pager containing the object type specified by the scope
      * @throws GitLabApiException if any exception occurs
      * @since GitLab 10.5
      */
-    public Pager<?> projectSearch(Object projectIdOrPath, ProjectSearchScope scope, String search, int itemsPerPage) throws GitLabApiException {
+    public Pager<?> projectSearch(Object projectIdOrPath, ProjectSearchScope scope, String search, String ref, int itemsPerPage) throws GitLabApiException {
```
gmessner, line 289: W still need a method that does not take a ref:
```
 public Pager<?> projectSearch(Object projectIdOrPath, ProjectSearchScope scope, String search, int itemsPerPage) throws GitLabApiException
```

ribardo0: Hi @gmessner, 

Effectively I should have kept the original method in addition to the new one (as I did for the 2 other methods). I just added a new commit to correct that.

Olivier

gmessner:
@ribardo0 This has been merged.  We plan on doing a new release on 4/4/2020, will let you know when it is available.

PR #549, opened 2020-04-06
Added OSGi Support

### LINKED ISSUE #548: Provide OSGi Support
Unfortunately the jar does not provide any meaningful Manifest, that would enable it for OSGi users. I will provide a simple PR in a Minute for that.

If you are not firm with OSGI, the solution is quite simple. I will add the [bnd-maven-plugin](https://github.com/bndtools/bnd/tree/master/maven/bnd-maven-plugin), which will scann the classes and generate the necessary Manifest. Nothing else will be required and should not take up any maintenance.

juergen-albert:
*fixes #548
* adds bnd-maven-plugin and configures maven jar plugin to pick up the
generated manifest

Signed-off-by: Juergen Albert <j.albert@data-in-motion.biz>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,6 +56,7 @@
 		<systemRules.version>1.19.0</systemRules.version>
 
 		<changelog-lib.version>1.59</changelog-lib.version>
+		<bnd.version>5.0.1</bnd.version>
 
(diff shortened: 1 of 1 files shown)


gmessner:
@juergen-albert Thanks for the contribution.

This has been merged and released in 4.14.26

Issue #552, opened 2020-04-16
Jobs returned by the Runners api include project information


Gaff:
I'm trying to use this API: https://docs.gitlab.com/ee/api/runners.html#list-runners-jobs

However the Job objects returned don't include the `project` information. Would it be possible to add this?

ribardo0:
Hi,
I just added a pull request to fix this issue.
Olivier

gmessner:
@Gaff  This has been added and released in 4.14,27, please verify and close this issue when you have time.

Gaff:
Sorry, missed this. Looks great - thanks!!

PR #553, opened 2020-04-17
Include Project in RunnersAPI.getJobs* (Issue #552)

ribardo0:
Hi

I added project attribute in Job class, this will work for RunnersAPI but may be confusing for JobAPI (that will never return Project Information but that uses the same class).

I also added withDuration missing method.

Olivier

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/Job.java
+++ b/src/main/java/org/gitlab4j/api/models/Job.java
@@ -300,7 +300,7 @@ public class Job {
         return this;
     }
 
-    public Job allowFailureManual(Boolean allowFailure) {
+    public Job withAllowFailure(Boolean allowFailure) {
         this.allowFailure = allowFailure;
         return this;
     }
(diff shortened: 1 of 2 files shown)


gmessner:
@ribardo0 Not a concern adding the ```project``` property.  Since you are down in the ```withXXX()```, there is one that needs fixed:
```
public Job allowFailureManual(Boolean allowFailure)
```
Should be:
```
public Job withAllowFailure(Boolean allowFailure)
```

Could you fix this as part of this PR?

ribardo0:
Hi @gmessner 

OK I will do it.
Should I remove it or only create a new method and deprecate the other one (for compatibility) ?

Olivier

gmessner:
@ribardo0 
Just rename it, I guarantee that no one is using it and with the remote chance someone is, they'll figure out.

ribardo0:
@gmessner 
As requested I changed method name to "withAllowFailure". I also included "project" object in Job.json

gmessner:
@ribardo0 Changes look good, merging.

Will have a release out in the next day or so.

PR #556, opened 2020-04-23
Update commit API to be conformant to spec

gsdatta:
Three updates in this PR.

**1. Fix url field in `Commit`**
The correct url field is `web_url`.

See: https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/lib/api/entities/commit.rb#L13 and https://docs.gitlab.com/ee/api/commits.html

**2. Remove `data` field from `Event`, which does not exist in the spec.**
See: https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/lib/api/entities/event.rb#L15

Additionally, the `push_data` field only exists in Push events, so update the test to reflect that.

**3. Remove `author` from `Commit`, which does not exist in the spec.**
v4: https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/lib/api/entities/commit.rb#L10
v3: https://gitlab.com/gitlab-org/gitlab-foss/-/blob/8-16-stable/doc/api/commits.md

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/Commit.java
+++ b/src/main/java/org/gitlab4j/api/models/Commit.java
@@ -8,6 +8,7 @@ import org.gitlab4j.api.utils.JacksonJson;
 
 public class Commit {
 
+    private Author author;
     private Date authoredDate;
     private String authorEmail;
     private String authorName;
@@ -23,8 +24,17 @@ public class Commit {
     private String status;
     private Date timestamp;
     private String title;
+    private String url;
     private String webUrl;
     private Pipeline lastPipeline;
+
+    public Author getAuthor() {
+        return author;
+    }
+
+    public void setAuthor(Author author) {
+        this.author = author;
+    }
 
     public Date getAuthoredDate() {
         return authoredDate;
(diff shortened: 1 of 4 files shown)


gmessner:
@gsdatta First off thank you for your contribution, however, it is not something that can be merged.

The reason is that this library does not only support the most current version but all versions of GitLab that are doing the v4 API, which goes back 3 or so years.  All the properties that are being removed will break the use of this library for many users that are not on the most current version of GitLab.

You did identify one change that is an addition, that would be the addition of the ```web_url``` in the ```Commit``` class.  The ```url``` property must remain because that is in earlier versions of GitLab.

To accept this MR you'll need to add back in the removed properties and add ```web_url``` to the ```Commit``` class meaning ```url``` must remain in place.

gsdatta:
Had a feeling that was the case :)

Just reverted those changes, but kept the changes to `commit.json` and `event.json` to make those tests more robust.

gsdatta:
Hi @gmessner, wanted to check whether the changes addressed your feedback or if further fixes were needed. Thanks!

gmessner:
@gsdatta Thanks for making the changes.  I will have a release out in the next day or so.

PR #558, opened 2020-04-27
Fix casing of get/set webUrl in Commit

gsdatta:
Noticed an error in the previous patch #556 - fixing the casing of the methods I added.

Sorry about the miss!


gmessner:
@gsdatta Easy one to miss because the JSON will marshal and unmarshal just fine.  I've merged and will be releasing on 5/2

PR #570, opened 2020-05-06
Expose method to update project deploy key

adamMo:


src/main/java/org/gitlab4j/api/DeployKeysApi.java
```diff
@@ -192,6 +192,24 @@ public DeployKey addDeployKey(Object projectIdOrPath, String title, String key,
...
+     * Updates an existing project deploy key.
+     *
+     * <pre><code>GitLab Endpoint: PUT /projects/:id/deploy_keys/:key_id</code></pre>
+     *
+     * @param projectIdOrPath the project in the form of an Integer(ID), String(path), or Project instance
+     * @param deployKey a Deploy key instance with updated properties
+     * @return an updated DeployKey instance
+     * @throws GitLabApiException if any exception occurs
+     */
+    public DeployKey updateDeployKey(Object projectIdOrPath, DeployKey deployKey) throws GitLabApiException {
```
gmessner, line 205: @adamMo Since only ```title``` and ```canPush``` properties can be updated and both are optional I would prefer  thta this method take parameters as follows:
```
public DeployKey updateDeployKey(Object projectIdOrPath, Integer deployKeyId, String title, Boolean canPush)
```
Sending the entire ```DeployKey``` instance will be sending information that is not needed.  If you create a ```DeployKey``` instance and populate it with just the values of ```title``` and ```canPush``` which either can be null will be better.

adamMo: I made it in two methods now, let me know if it's acceptable for you.

src/main/java/org/gitlab4j/api/DeployKeysApi.java
```diff
@@ -192,6 +192,50 @@ public DeployKey addDeployKey(Object projectIdOrPath, String title, String key,
...
+     * <pre><code>GitLab Endpoint: PUT /projects/:id/deploy_keys/:key_id</code></pre>
+     *
+     * @param projectIdOrPath the project in the form of an Integer(ID), String(path), or Project instance
+     * @param canPush can deploy key push to the project's repository, optional
+     * @param deployKeyId the ID of the deploy key to update
+     * @param title the new deploy key's title, required
+     * @return an updated DeployKey instance
+     * @throws GitLabApiException if any exception occurs
+     */
+    public DeployKey updateDeployKey(Object projectIdOrPath, Integer deployKeyId, Boolean canPush, String title) throws GitLabApiException {
```
gmessner, line 228: Please order the parameters in the order requested:
```java
public DeployKey updateDeployKey(Object projectIdOrPath, Integer deployKeyId, String title, Boolean canPush)
```
This is to make it match the GitLab API documentation at:
https://docs.gitlab.com/ee/api/deploy_keys.html#update-deploy-key

And make sure to list ```title``` as optional and use the word "updated" instead of "new", oddly the documentation at the above link has it wrong.

src/main/java/org/gitlab4j/api/DeployKeysApi.java
gmessner: @adamMo Thanks for the update, however I'd like to remove the method that takes a ```DeployKey``` instance.  The pattern for this library is to pass objects in only when there are many parameters.

adamMo: Ok, it's removed.

src/main/java/org/gitlab4j/api/DeployKeysApi.java
adamMo: Sure, changed.

gmessner:
@adamMo Thanks again for the contribution.  I'll try and get a release out in the next couple of days.

adamMo:
@gmessner Did you think about automatic releases using GitHub Actions or some other platform? I see that there are tests depending on gitlab docker container... I have some experience in setting up such pipelines in Gitlab CI, didn't try Actions. What do you think?

gmessner:
@adamMo Thanks for the offer, but I have no desire to "fully" automate the release of GitLab4J-API.  The actual release is automated on Jenkins, however there are many things that determine when to do a release, and I do not wish to release just because a PR was merged or something was pushed to the repo, this could cause too many releases, and many times adjustments are needed to merged code before it is ready to be released.

Currently, I release twice monthly unless there is an urgent need of a releasing because of a bug fix or major new functionality. 

The reason I said "I'll try and get a release out in the next couple of days." was because I had something I was working on that was also going to be included in the release.  That being said, the release was actually done yesterday and 4.14.29 is already available with your addition in it, just had not got around to notifying you.

PR #573, opened 2020-05-14
Extended "Get a single pipeline schedule" endpoint with Variables

mashinarius:
Extended "Get a single pipeline schedule" endpoint with Variables attribute.

In accordance to documentation 
https://docs.gitlab.com/ee/api/pipeline_schedules.html#get-a-single-pipeline-schedule
"Get the pipeline schedule of a project" request also returns the list of variables, that were added to Pipeline Scheduler.


mashinarius:
> @mashinarius No need to add another class just add the `variables` property and getter and setter to the existing `PipelineSchedule` class.

Thank you for your review.
Just one clarification. 
In case we want to "get all pipeline schedules" via "GET /projects/:id/pipeline_schedules"
Gitlab will return all base attributes except Variables.
https://docs.gitlab.com/ee/api/pipeline_schedules.html#get-all-pipeline-schedules

And having Variables in PipelineSchedule class could lead the developer to the false sense that there are no Variables in the pipeline schedule.

What do you think about this?

gmessner:
@mashinarius 
Have no concerns about the developer on this.  This library has a 1-to-1 relationship with the GitLab API, and there are literally hundreds of this sort of thing where depending on what endpoint is being called there is a variation of what is returned.  If we created a class for every variation we would have many hundreds of model classes actually making it more difficult to consume.

mashinarius:
Ok, clear
An additional class was removed, variables added to original PipelineSchedule class

gmessner:
@mashinarius Thanks for the updates.  I will merge and tomorrows release will include this.  Will let you know when the release is available.

gmessner:
@mashinarius GitLab4J-API 4.14.30 is now available with your updates.

mashinarius:
@gmessner  thank you so much

PR #590, opened 2020-06-29
Remove UserListSerializer and UserListDeserializer from approvedBy field in ApprovalRule (fixes #589)

### LINKED ISSUE #589: UserListSerializer and UserListDeserializer should not be used on the approvedBy field of ApprovalRule
**Description of Problem**
`UserListSerializer` and `UserListDeserializer` are being applied to the `approvedBy` field of `ApprovalRule`. This is not correct, because this field is a normal list of users and does not need the special serializer and deserializer. You can see this is true by looking at the example from [GitLab's documentation](https://docs.gitlab.com/ee/api/merge_request_approvals.html#get-the-approval-state-of-merge-requests):
```
{
  "approval_rules_overwritten": true,
  "rules": [
    {
      "id": 1,
      "name": "Ruby",
      "rule_type": "regular",
      "eligible_approvers": [
        {
          "id": 4,
          "name": "John Doe",
          "username": "jdoe",
          "state": "active",
          "avatar_url": "https://www.gravatar.com/avatar/0?s=80&d=identicon",
          "web_url": "http://localhost/jdoe"
        }
      ],
      "approvals_required": 2,
      "users": [
        {
          "id": 4,
          "name": "John Doe",
          "username": "jdoe",
          "state": "active",
          "avatar_url": "https://www.gravatar.com/avatar/0?s=80&d=identicon",
          "web_url": "http://localhost/jdoe"
        }
      ],
      "groups": [],
      "contains_hidden_groups": false,
      "approved_by": [
        {
          "id": 4,
          "name": "John Doe",
          "username": "jdoe",
          "state": "active",
          "avatar_url": "https://www.gravatar.com/avatar/0?s=80&d=identicon",
          "web_url": "http://localhost/jdoe"
        }
      ],
      "source_rule": null,
      "approved": true,
      "overridden": false
    }
  ]
}
```
Note that `approved_by` is the exact same as `users`.

`UserListSerializer` and `UserListDeserializer` cause exceptions when trying to interact with merge request approval state JSON because of this mismatch.

**Suggested Fix**
The serializer and deserializer annotations should be removed from the `approvedBy` field of `ApprovalRule`.

tyates-indeed:
Remove UserListSerializer and UserListDeserializer from approvedBy field in ApprovalRule (fixes #589)

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/ApprovalRule.java
+++ b/src/main/java/org/gitlab4j/api/models/ApprovalRule.java
@@ -18,9 +18,6 @@ public class ApprovalRule {
     private List<User> users;
     private List<Group> groups;
     private Boolean containsHiddenGroups;
-
-    @JsonSerialize(using = JacksonJson.UserListSerializer.class)
-    @JsonDeserialize(using = JacksonJson.UserListDeserializer.class)
     private List<User> approvedBy;
     private Boolean approved;
 
--- a/src/main/java/org/gitlab4j/api/models/ApprovalRule.java
+++ b/src/main/java/org/gitlab4j/api/models/ApprovalRule.java
@@ -1,11 +1,8 @@
 package org.gitlab4j.api.models;
 
-import java.util.List;
-
 import org.gitlab4j.api.utils.JacksonJson;
 
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.util.List;
 
 public class ApprovalRule {
 
--- a/src/test/java/org/gitlab4j/api/TestGitLabApiBeans.java
+++ b/src/test/java/org/gitlab4j/api/TestGitLabApiBeans.java
@@ -28,6 +28,7 @@ import org.gitlab4j.api.models.AccessRequest;
 import org.gitlab4j.api.models.Application;
 import org.gitlab4j.api.models.ApplicationSettings;
 import org.gitlab4j.api.models.ApprovalRule;
+import org.gitlab4j.api.models.ApprovalState;
 import org.gitlab4j.api.models.ArtifactsFile;
 import org.gitlab4j.api.models.AuditEvent;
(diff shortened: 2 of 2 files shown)


src/main/java/org/gitlab4j/api/models/ApprovalRule.java
```diff
@@ -18,9 +18,6 @@
     private List<User> users;
     private List<Group> groups;
     private Boolean containsHiddenGroups;
-
-    @JsonSerialize(using = JacksonJson.UserListSerializer.class)
-    @JsonDeserialize(using = JacksonJson.UserListDeserializer.class)
```
derekyyyan, line 23: Should also remove the unused imports.

tyates-indeed: Yes, good catch. I have optimized imports with IntelliJ now.

gmessner:
@tyates-indeed  and @derekyyyan 
The problem is that this will break the [Merge Request-level MR approvals](https://docs.gitlab.com/ee/api/merge_request_approvals.html#merge-request-level-mr-approvals)  

It is possible the documentation is incorrect, if it is correct then a bug issue should be filed with GitLab to make it consistent.

tyates-indeed:
@gmessner all the unit tests passed which I assumed meant this would be okay. Are there not existing test cases for Merge Request-level MR approvals?

gmessner:
@tyates-indeed Unfortunately, there are no existing unit tests for the Merge Request-level MR approvals which is why the unit tests passed.  I'll add unit tests for the Merge Request-level MR approvals  and then we figure out how to best address this inconsistency in the GitLab API

gmessner:
@tyates-indeed After a thorough examination of the GitLab approval related API endpoints I have determined that this change is actually correct,  The calls that require the UserListerializer/Deserializer return an instance of ```MergeRequest``` which is properly configured, so this change to ```ApprovalRule``` is correct. I will be merging and releasing today.

gmessner:
@tyates-indeed GitLab4J-API Release 4.15.3 has been published and is now available.

PR #600, opened 2020-07-31
added buildCoverageRegex and buildGitStrategy to Project

kvakos:
adds two attributes to Project


src/main/java/org/gitlab4j/api/models/BuildGitStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+    @JsonValue
+    public String toValue() {
+        return (enumHelper.toString(this));
+    }
+
+    @Override
+    public String toString() {
+        return (enumHelper.toString(this));
+    }
+}
```
gmessner, line 53: @kvakos Unless an enum is shared by multiple models we usually will not create a stand-alone class for it, please move this into ```org.gitlab4j.api.Constants``` (preferred) or directly into ```org.gitlab4j.api.models.Project```

kvakos: BuildGitStrategy is now inner class  of Project

kvakos: I changed to preferred implementation.

gmessner:
@kvakos Release 4.15.4 is now available with this functionality.

PR #607, opened 2020-08-21
Add search_namespaces to GroupProjectsFilter

marcojahn:
adds the new search_namespace boolean flag when using search(), to search within namespaces


gmessner:
@marcojahn Thanks for the contribution, however, ```search_namespaces``` is not a parameter when searching for projects using the ```GroupApi```. See https://docs.gitlab.com/ce/api/groups.html#list-a-groups-projects

PR #655, opened 2021-01-28
feat: add support for GitLab merge request reviewers

LupusUmbrae:
Add the reviewer field to merge requests

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/MergeRequest.java
+++ b/src/main/java/org/gitlab4j/api/models/MergeRequest.java
@@ -15,6 +15,7 @@ public class MergeRequest {
     private Integer approvalsBeforeMerge;
     private Assignee assignee;
     private List<Assignee> assignees;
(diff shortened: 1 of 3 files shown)


src/main/java/org/gitlab4j/api/models/MergeRequest.java
```diff
@@ -15,6 +15,7 @@
     private Integer approvalsBeforeMerge;
     private Assignee assignee;
     private List<Assignee> assignees;
+    private List<Assignee> reviewers;
```
eedzjee, line 18: Probably a good thing to define a new class `Reviewer` that extends from `AbstractUser` for this?

LupusUmbrae: I thought about it, it just felt redundant to create an identical class but would be more in keeping with the rest of the system

jabby (approved):

PR #685, opened 2021-04-09
Fix tests relying on "X-Total-Pages" and "X-Total" headers that have been removed

maltem-za:
This is part of a set of PRs with the purpose of fixing broken tests for the latest version of GitLab (13.10.2 at time of writing) and so setting the project up for ongoing development with a working suite of tests.

Also see #683 & #684 - with the changes from the three related PRs all tests are passing for me against a gitlab-ee Docker image.

More can be done to improve these tests, as noted in the comments.


jmini:
I suggest to review this when we have the possibility to run CI tests against a more recent version of GitLab (see #925)

Issue #653, opened 2021-01-18
no support for users/personal_access_tokens


msilvestre:
Is it possible to add support for endpoint: users/personal_access_tokens

Thank you.

sleiss:
This would be really helpful!

PR #686, opened 2021-04-09
feat: implement createPersonalAccessToken

maltem-za:
Add support for Personal Access Token creation (#653)

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/PersonalAccessToken.java
+++ b/src/main/java/org/gitlab4j/api/models/PersonalAccessToken.java
@@ -9,12 +9,12 @@ import org.gitlab4j.api.utils.JacksonJsonEnumHelper;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 
-public class ImpersonationToken {
+public class PersonalAccessToken {
 
-    /** Enum to specify the scope of an ImpersonationToken. */
+    /** Enum to specify the scope of a PersonalAccessToken. */
     public enum Scope {
 
-        API, READ_USER, READ_REPOSITORY, WRITE_REPOSITORY, READ_REGISTRY, SUDO;
+        API, READ_API, READ_USER, READ_REPOSITORY, WRITE_REPOSITORY, READ_REGISTRY, WRITE_REGISTRY, SUDO;
 
         private static JacksonJsonEnumHelper<Scope> enumHelper = new JacksonJsonEnumHelper<>(Scope.class);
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/gitlab4j/api/models/PersonalAccessToken.java
```diff
@@ -9,12 +9,12 @@
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 
-public class ImpersonationToken {
+public class PersonalAccessToken {
```
jmini, line 12: Renaming the class like this is a breaking change for the API consumer.

jmini:
@maltem-za since the change is breaking, would it be OK to do in our next major version `6.0.0` on the [`6.x` branch](https://github.com/gitlab4j/gitlab4j-api/tree/6.x)?

maltem-za:
Hi @jmini, yes of course, whatever works for you. I switched to my own fork anyway as I needed this at the time that I opened the PR (it's >2 years old now). You're right of course that it's a breaking change, but in my mind it's a pretty simple refactor on the consumer side.

I stopped opening new PRs because it didn't seem likely they'd get merged, but you may be interested in integrating some of the other changes I've been making in my fork (mostly keeping tests working for later GitLab versions, and more recently I implemented OAuth2 access token refresh [here](https://github.com/forTEXT/gitlab4j-api/commit/1b0938786a204661d23200c36913a639df55ba55)). Happy to help where I can, time permitting :)

jmini:
@maltem-za:
I have rebased your change on top of `origin/6.x` but I am not able to force push it on your branch `add-pats-support`.

Probably you did not check `[x] allowed edits from maintainers`

---

As for working on a fork or in the upstream project, this is always a dilemma.
I joined this repo recently to not have to maintain my own fork, but I understand your reasons (and 2 years ago there was not much activity on the project).

maltem-za:
@jmini Hmm, I can't find that option. Do I need to open a new PR?

I'd be happy to contribute more directly again if things start moving a bit faster, and I would also prefer to not have to maintain a fork. Unfortunately I won't be able to make the switch to Java 11+ that quickly, but I guess the plan is to run 5.x and 6.x in parallel for some time anyway. I'll keep an eye on how things progress. Thanks for picking this up!

jmini:
I have opened a new PR for this #963.

On the main branch (version `5.x.x`), my suggestion is to keep the old name `ImpersonationToken` and to do the renaming to `PersonalAccessToken` only with version `6.x.x`.

PR #689, opened 2021-04-12
Support larger file sizes by expanding the size type to a Long.

martin1yness:
Tried getting Job and got a deserialization error, I think this will resolve it.

Caused by: com.fasterxml.jackson.databind.JsonMappingException: Numeric value (3328850515) out of range of int (-2147483648 - 2147483647)
 at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 3915] (through reference chain: java.util.ArrayList[1]->org.gitlab4j.api.models.Job["artifacts_file"]->org.gitlab4j.api.models.ArtifactsFile["size"])


martin1yness:
Need same fix in the Artifact model on the size property.

Issue #631, opened 2020-11-04
getJobApi().getJobsForPipeline() does not support paging


theotherp:
Unfortunately in my case when calling `getJobApi().getJobsForPipeline()` gitlab returns more than 100 results. Paging support for this call would be great. I don't think there's any other way of calling this paged, right? I'd call the `get` method and do the rest myself but it's protected.

PR #692, opened 2021-04-15
Added Paging- and Stream-Support for Jobs in Pipeline.

opensource21:
Fixing #631


opensource21:
The Build must be broken before. There fails a check of class I didn't touch.

jabby (approved):

PR #694, opened 2021-04-16
Add support for ci_config_path and ci_forward_deployment_enabled properties to Projects API

filipowm:
Currently Projects API is missing `ci_config_path` (create/read/update) and `ci_forward_deployment_enabled` (read/update), thus I'm adding this.


lastthyme (approved):

faymard:
Hi @filipowm ,

I'm also currently working on a fix to add attributes to some APIs (mostly Groups, Users and Projects). 

I'd just want to add that `ci_forward_development_enabled` is not supported in versions prior to GitLab 13.0. As some companies (and the integration testing of this project) still rely on 12.x, I thought it would be cool to add this as a side-note. It is not a breaking change, though.

faymard (approved):

serega6531:
I'd also love to see this in the build. Can we get this merged?

PR #698, opened 2021-05-04
Add UserApi.getMemberships() to list projects and groups a user is member of

bwaldvogel:


src/main/java/org/gitlab4j/api/models/MembershipSourceType.java
```diff
@@ -0,0 +1,8 @@
+package org.gitlab4j.api.models;
+
+public enum MembershipSourceType {
```
jabby, line 3: In this project all enum names should be capitalized. Can you modify this enum like [PackageType](https://github.com/gitlab4j/gitlab4j-api/blob/master/src/main/java/org/gitlab4j/api/models/PackageType.java)

jabby (changes requested):
Thanks for your contribution. It needs a little modification. Can you take care of this one?

src/main/java/org/gitlab4j/api/models/MembershipSourceType.java
bwaldvogel: Oh, sorry. Didn’t notice this coding convention.
I’ve adjusted the PR as you suggested.

jabby: Thanks

jabby (approved):

bwaldvogel:
I rebased my PR in the hope that the tests will succeed now. (Locally it succeeded)

PR #721, opened 2021-06-26
Fix #621 : Add deployment API

### LINKED ISSUE #621: Deployments API support
It seems GitLab4J lacks [Deployments API](https://docs.gitlab.com/ee/api/deployments.html) support.

jabby:


jabby:
Need some tests. Will add them in next commit.

PR #747, opened 2021-08-24
Change position to be Double instead of Integer

### LINKED ISSUE #669: Position model x and y should be floats
GitLab changed the data type of the `x` and `y` parameters of Position in v13.7 but did not update their documentation.

I opened an issue with them and they have now updated their docs to correctly list the data type for the `x` and `y` coordinates of position as `float`: https://gitlab.com/gitlab-org/gitlab/-/issues/324201

gitlab4j should be updated to support these new data types as current behavior with v13.7+ of GitLab causes exceptions when trying to parse image diff notes since the float values for `x` and `y` cannot be converted to the Integer type that gitlab4j expects.

krische:
This should fix #669

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/Position.java
+++ b/src/main/java/org/gitlab4j/api/models/Position.java
@@ -42,8 +42,8 @@ public class Position {
 
     private Integer width;
     private Integer height;
-    private Integer x;
-    private Integer y;
+    private Double x;
+    private Double y;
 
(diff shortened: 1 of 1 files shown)


jabby (approved):

krische:
@jabby Are you able to deliver a release with all of the PR's merged recently?

jabby:
@krische I will deliver a release tonight.

PR #752, opened 2021-09-03
Add 'runners_token' attribute to group model.

Poeschl:
This PR adds the group registration token for gitlab runners to the group model.

See https://docs.gitlab.com/ee/api/groups.html#details-of-a-group


src/main/java/org/gitlab4j/api/models/Group.java
```diff
@@ -295,6 +303,11 @@ public Group withSharedProjects(List<Project> sharedProjects) {
         return this;
     }
 
+    public Group withRunnersToken(String runnersToken) {
```
jabby, line 306: I don't see in [Groups documentation](https://docs.gitlab.com/ee/api/groups.html) a reference to a creation of a group with a specified `runners_token`. Are you sure this method is useful?

Poeschl: Since the token is only returned from the api and can't be set over the api, so I will remove the builder method.

PR #768, opened 2021-10-05
Fix #767 : Adding merge request note award functions and renaming issue functions

### LINKED ISSUE #767: Adding award emoji to merge request note is not available

didacrarbones:


dren-dk (approved):
This seems like a worthwhile improvement.

The method renames are needed to be more correct, but upgrading ought to be trivial.

jabby:
@dren-dk I agree with you. I will keep older methods for retro-compatibility. Old methods will be flagged as `Deprecated` and I will finish the renaming as the current code does not compile (see deleteNoteAwardEmoji at the end of the file)

I think @didacrarbones don't push latest version of the patchset but I value the work done. 

BTW @dren-dk thanks for the review on all PR. I appreciate a lot.

jabby:
PR replace by #812 

@didacrarbones thansk for your work on this.

PR #776, opened 2021-10-16
Upload: accept an InputStream besides a File

alb-i986:
This makes it possible to stream data coming from another server straight into the POST request to Gitlab, saving an extra, expensive, and useless IO operation on disk.

Use case: integrating Gitlab with another defect management system.

### DIFF
--- a/src/main/java/org/gitlab4j/api/GitLabApiClient.java
+++ b/src/main/java/org/gitlab4j/api/GitLabApiClient.java
@@ -2,6 +2,7 @@ package org.gitlab4j.api;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.Socket;
 import java.net.URL;
 import java.security.GeneralSecurityException;
@@ -12,7 +13,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLEngine;
@@ -29,7 +29,6 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.StreamingOutput;
-
 import org.gitlab4j.api.Constants.TokenType;
 import org.gitlab4j.api.GitLabApi.ApiVersion;
 import org.gitlab4j.api.utils.JacksonJson;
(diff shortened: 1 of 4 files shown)


src/main/java/org/gitlab4j/api/GitLabApiClient.java
```diff
@@ -600,23 +600,37 @@ protected Response upload(String name, File fileToUpload, String mediaTypeString
      * @throws IOException if an error occurs while constructing the URL
      */
     protected Response upload(String name, File fileToUpload, String mediaTypeString, Form formData, URL url) throws IOException {
+        MediaType mediaType = (mediaTypeString != null ? MediaType.valueOf(mediaTypeString) : null);
+        FileDataBodyPart filePart = mediaType != null ?
```
jabby, line 604: After checking in `FileDataBodyPart` constructors, you don't need to test if mediaType is null or not. In fact the constructor used line 606 call the one used line 605 with null in MediaType. Please remove the ternary operator.

src/main/java/org/gitlab4j/api/GitLabApiClient.java
```diff
@@ -600,23 +600,37 @@ protected Response upload(String name, File fileToUpload, String mediaTypeString
...
 
+    protected Response upload(String name, InputStream inputStream, String filename, String mediaTypeString, Object... pathArgs) throws IOException {
+        URL url = getApiUrl(pathArgs);
+        return (upload(name, inputStream, filename, mediaTypeString, null, url));
+    }
+
+    protected Response upload(String name, InputStream inputStream, String filename, String mediaTypeString, Form formData, URL url) throws IOException {
         MediaType mediaType = (mediaTypeString != null ? MediaType.valueOf(mediaTypeString) : null);
-        try (FormDataMultiPart multiPart = new FormDataMultiPart()) {
+        StreamDataBodyPart streamDataBodyPart = mediaType != null ?
```
jabby, line 617: After checking in `StreamDataBodyPart` constructors, you don't need to test if mediaType is null or not. In fact the constructor used line 619 call the one used line 618 with null in MediaType. Please remove the ternary operator.

jabby (changes requested):
Thank you for your PR and sorry for taking so much time before reviewing your work.

src/main/java/org/gitlab4j/api/GitLabApiClient.java
alb-i986: Oh yes, I see.
Nice, we can rely on `FileDataBodyPart`'s capability of detecting the media type from a `File`.

The method parameter `mediaTypeString` should be removed then.
Since it's declared as `protected` it is exposed to clients, so we may not remove it until we get to a major version, isn't it?
I'll just update the javadocs for now.

src/main/java/org/gitlab4j/api/GitLabApiClient.java
alb-i986: In this case it's not really the same thing.
As `StreamDataBodyPart`'s javadoc states:

> It sets the InputStream as a body part with the default MediaType.APPLICATION_OCTET_STREAM_TYPE (if not specified by the user).
> Note: The MIME type of the entity cannot be automatically predicted as in case of FileDataBodyPart.

[Gitlab docs](https://docs.gitlab.com/ee/api/projects.html#upload-a-file) do not help clarifying.
But I guess it's not a problem, I think Gitlab REST API doesn't care about the media type.
What do you think @jabby?

jabby (approved):

alb-i986:
Thanks a lot!

Cheers

PR #786, opened 2021-11-24
Feature: Replace auth token with supplier

rtorsten:
Use case: oauth2 access limited time tokens which should be updated on regular basis. Supplier allows us to move refresh logic outside client.
Use specified setter to set up supplier after gitlab api object creation.


dren-dk (approved):
Very nice and minimal improvement with backward compatible API .

jabby (approved):

PR #796, opened 2022-02-14
Fixed MembershipSourceType enum, UserApi.getMemberships() and implemented KeysApi

dren-dk:
* User Memberships could not be used due to the MembershipSourceType always being null in the Membership, because Gitlab returns "Namespace" and "Project" rather than "NAMESPACE" and "PROJECT". The fix is to introduce a name field on the enum to have a place to store that string and make it available to users and to uppercase the string before trying to parse it as an enum name.

* Fixed the UserApi.getMemberships() method so it returns all the memberships and added a Pager based method too for lazy fetching.

* Added the KeysApi needed to look up users by ssh key fingerprint.

### DIFF
--- a/src/main/java/org/gitlab4j/api/GitLabApi.java
+++ b/src/main/java/org/gitlab4j/api/GitLabApi.java
@@ -95,6 +95,7 @@ public class GitLabApi implements AutoCloseable {
     private TodosApi todosApi;
     private UserApi userApi;
     private WikisApi wikisApi;
+    private KeysApi keysApi;
 
     /**
      * Get the GitLab4J shared Logger instance.
@@ -1673,6 +1674,21 @@ public class GitLabApi implements AutoCloseable {
 
         return wikisApi;
     }
+
+    /**
+     * Gets the KeysApi instance owned by this GitLabApi instance. The KeysApi is used to look up users by their ssh key signatures
+     *
+     * @return the KeysApi instance owned by this GitLabApi instance
+     */
+    public KeysApi getKeysAPI() {
+        synchronized (this) {
+            if (keysApi == null) {
+                keysApi = new KeysApi(this);
+            }
+        }
+        return keysApi;
+    }
+
 
     /**
      * Create and return an Optional instance associated with a GitLabApiException.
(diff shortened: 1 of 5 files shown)


src/main/java/org/gitlab4j/api/models/MembershipSourceType.java
```diff
@@ -1,31 +1,38 @@
...
+    PROJECT("Project"),
 
     /** Representing a group */
-    NAMESPACE;
+    NAMESPACE("Namespace");
 
-    private static JacksonJsonEnumHelper<MembershipSourceType> enumHelper = new JacksonJsonEnumHelper<>(MembershipSourceType.class);
+    public final String name;
+
+    MembershipSourceType(String name) {
```
jabby, line 16: We don't need to add an attribute name here. We should use a different JacksonJsonEnumHelper constructor. See `AwardableType` for example. 
Can you take care of this.

src/main/java/org/gitlab4j/api/GitLabApi.java
```diff
@@ -1674,6 +1675,21 @@ public WikisApi getWikisApi() {
         return wikisApi;
     }
 
+    /**
+     * Gets the KeysApi instance owned by this GitLabApi instance. The KeysApi is used to look up users by their ssh key signatures
+     *
+     * @return the KeysApi instance owned by this GitLabApi instance
+     */
+    public KeysApi getKeysAPI() {
```
jabby, line 1683: Can you split KeysApi implementation and MembershipSourceType fix in 2 differents PR?

jabby (changes requested):
Hello @dren-dk 

Thanks for your work. Can you take care of my review. 

have a nice day.

src/main/java/org/gitlab4j/api/models/MembershipSourceType.java
dren-dk: Done: https://github.com/gitlab4j/gitlab4j-api/pull/816

src/main/java/org/gitlab4j/api/GitLabApi.java
dren-dk: Done: https://github.com/gitlab4j/gitlab4j-api/pull/814

dren-dk:
Replaced this PR with 3 new ones:
https://github.com/gitlab4j/gitlab4j-api/pull/814
https://github.com/gitlab4j/gitlab4j-api/pull/815
https://github.com/gitlab4j/gitlab4j-api/pull/816

PR #797, opened 2022-02-15
Add top_level_only parameter to GroupFilter

Djcd:
Hi,

as described in https://docs.gitlab.com/ee/api/groups.html there is a parameter filter for only top level groups. I added the parameter to the GroupFilter


dren-dk (approved):
Worthwhile improvement IMHO.

jabby (approved):

PR #815, opened 2022-03-09
Actually fetch all the memberships with UserAPI.getMemberships() and …

dren-dk:
…also provide a pager based method for lazy fetching


jabby (approved):
Thanks for your work. I will modify a little bit your implementation. 

- add `itemsPerPage` attribute for the getMembershipsPager
- rename getMembershipsPager to getMemberships
- use `.all()` on pager instead of `.stream().collect(Collectors.toList());`

PR #822, opened 2022-03-14
Adding getContributors sort by asc / desc support

maddymanu:
Hey guys!
I made a simple addition to the `RepositoryApi`.

I wasn't able to find contribution guidelines for this repo, but I'd be happy to make any requested changes. 

Change Log
- New `getContributors` API, to sort by `desc` or `asc`.

Documentation for this API: https://docs.gitlab.com/ee/api/repositories.html#contributors

### DIFF
--- a/src/main/java/org/gitlab4j/api/RepositoryApi.java
+++ b/src/main/java/org/gitlab4j/api/RepositoryApi.java
@@ -664,6 +664,33 @@ public class RepositoryApi extends AbstractApi {
     }
 
     /**
+     * Get a list of contributors from a project and in the specified page range, sorted by specified param.
+     *
+     * <pre><code>GitLab Endpoint: GET /projects/:id/repository/contributors</code></pre>
+     *
+     * @param projectIdOrPath the project in the form of an Long(ID), String(path), or Project instance
+     * @param page the page to get
+     * @param perPage the number of projects per page
+     * @param sort optional param to sort the list of contributors by
+     * @return a List containing the contributors for the specified project ID
+     * @throws GitLabApiException if any exception occurs
+     */
+    public List<Contributor> getContributors(Object projectIdOrPath, int page, int perPage, String sort) throws GitLabApiException {
+        if (sort != null && !(sort.equals("asc") || sort.equals("desc")) ) {
+            throw new RuntimeException("Sort must be asc or desc");
+        }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/gitlab4j/api/RepositoryApi.java
```diff
@@ -663,6 +663,33 @@ public CompareResults compare(Object projectIdOrPath, String from, String to) th
...
+     * <pre><code>GitLab Endpoint: GET /projects/:id/repository/contributors</code></pre>
+     *
+     * @param projectIdOrPath the project in the form of an Long(ID), String(path), or Project instance
+     * @param page the page to get
+     * @param perPage the number of projects per page
+     * @param sort optional param to sort the list of contributors by
+     * @return a List containing the contributors for the specified project ID
+     * @throws GitLabApiException if any exception occurs
+     */
+    public List<Contributor> getContributors(Object projectIdOrPath, int page, int perPage, String sort) throws GitLabApiException {
```
jabby, line 678: - Add all possibilities of order by. You can add a new enum for this. It will help anybody who wants to sort contributors by  name or email (default one is by number of commits).
- Use `SortOrder sortOrder` instead of `String sort`. It will enable compilation check instead of runtime check like you do. 

See the following classes as an example. 

https://github.com/gitlab4j/gitlab4j-api/blob/0947d1fa324a0551ab9119381c8373351c7b13e3/src/main/java/org/gitlab4j/api/Constants.java#L89-L109 
and https://github.com/gitlab4j/gitlab4j-api/blob/master/src/main/java/org/gitlab4j/api/EpicsApi.java#L107-L134

jabby (changes requested):
Hello, 

Thanks for your PR. 
I suggest some improvents before merging. Can you take care of them?

Have a nice day.

src/main/java/org/gitlab4j/api/RepositoryApi.java
maddymanu: Hey @jabby !! 
thanks for the quick feedback!

I just addressed all of it. Lemme know if this looks good :)

jabby: Looks good to me

jabby (approved):

PR #870, opened 2022-10-08
Add method to update code_owner_approval_required flag to Protected Branches API

### LINKED ISSUE #869: Protected Branches API doesn't support updating code_owner_approval_required on existing protected branch
The standard GitLab API supports [PATCH'ing an existing protected branch](https://docs.gitlab.com/ee/api/protected_branches.html#require-code-owner-approvals-for-a-single-branch) to set the `code_owner_approval_required` flag. Otherwise the current gitlab4j appears to require unprotecting then protecting the branch with the new setting (which would leave the branch unprotected for some window of time, which you don't want).

For example, curling the API directly you could do something like this:
```
curl --request PATCH --header "PRIVATE-TOKEN: <your_access_token>" "https://gitlab.example.com/api/v4/projects/5/protected_branches/feature-branch?code_owner_approval_required=true"
```

MikeWillCook:
Adds basic support for updating the `code_owner_approval_required` flag on an existing protected branch. See the [GitLab API docs](https://docs.gitlab.com/ee/api/protected_branches.html#require-code-owner-approvals-for-a-single-branch) for intended usage.

Two notes for feedback:
1. The HTTP "PATCH" method isn't supported in the current client so this uses the `X-HTTP-Method-Override: PATCH` header as a workaround. This wires it thru AbstractApi to GitLabApiClient where it's turned into a POST with the override header. Anyone recommend another way?
2. Removed the integration test as it's a Premium or higher feature so can't be tested on the Community Edition image. (EDIT: Added this test back with handling for current version.)

Fixes #869

### DIFF
--- a/src/main/java/org/gitlab4j/api/GitLabApiClient.java
+++ b/src/main/java/org/gitlab4j/api/GitLabApiClient.java
@@ -471,6 +471,35 @@ public class GitLabApiClient implements AutoCloseable {
     }
 
     /**
+     * Perform an HTTP PATCH call with the specified query parameters and path objects, returning
+     * a ClientResponse instance with the data returned from the endpoint.
+     *
+     * @param queryParams multivalue map of request parameters
+     * @param pathArgs variable list of arguments used to build the URI
+     * @return a ClientResponse instance with the data returned from the endpoint
+     * @throws IOException if an error occurs while constructing the URL
+     */
+    protected Response patch(MultivaluedMap<String, String> queryParams, Object... pathArgs) throws IOException {
+        URL url = getApiUrl(pathArgs);
+        return (patch(queryParams, url));
+    }
+
+    /**
+     * Perform an HTTP PATCH call with the specified query parameters and URL, returning
+     * a ClientResponse instance with the data returned from the endpoint.
+     *
+     * @param queryParams multivalue map of request parameters
+     * @param url the fully formed path to the GitLab API endpoint
+     * @return a ClientResponse instance with the data returned from the endpoint
+     */
+    protected Response patch(MultivaluedMap<String, String> queryParams, URL url) {
+        Entity<?> empty = Entity.text("");
+        // use "X-HTTP-Method-Override" header on POST to override to unsupported PATCH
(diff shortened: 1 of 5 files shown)


src/main/java/org/gitlab4j/api/GitLabApiClient.java
```diff
@@ -470,6 +470,35 @@ protected Response head(MultivaluedMap<String, String> queryParams, URL url) {
...
+     * a ClientResponse instance with the data returned from the endpoint.
+     *
+     * @param queryParams multivalue map of request parameters
+     * @param url the fully formed path to the GitLab API endpoint
+     * @return a ClientResponse instance with the data returned from the endpoint
+     */
+    protected Response patch(MultivaluedMap<String, String> queryParams, URL url) {
+        Entity<?> empty = Entity.text("");
+        // use "X-HTTP-Method-Override" header on POST to override to unsupported PATCH
+        return (invocation(url, queryParams)
```
jabby, line 498: Hello thanks for your PR. 

I did not check if it works, but I think you can use `invocation(url, queryParams).method("PATCH", empty);`

Can you try with my proposal ?

MikeWillCook: Thanks for the suggestion, @jabby. In fact I tried that first (after not finding a native patch() method) and found it doesn't work with the current client, and instead returns the following exception:
```
org.gitlab4j.api.GitLabApiException: java.net.ProtocolException: Invalid HTTP method: PATCH
```
There is some [discussion](https://stackoverflow.com/questions/22355235/patch-request-using-jersey-client) of other workarounds for the [underlying bug](https://bugs.openjdk.org/browse/JDK-8207840), but this approach seemed sufficient and less problematic to me. I'm open to pushback on that.

Also, to help in the future and at least exercise the method, I've added a test handling the current tested version with comments on what should happen with Premium.

MikeWillCook: Any other questions on this, @jabby ?

jabby (approved):

PR #885, opened 2022-12-13
Deprecate MergeRequest#mergeStatus

### LINKED ISSUE #880: Deprecation of the `merge_status` API field
Gitlab 15.6 deprecates the `merge_status` field in their merge request API - planned removal is for Gitlab 16.0 (2023-05-22): https://docs.gitlab.com/ee/update/deprecations.html#merge_status-api-field

They are introducing a new field `detailed_merge_status` to replace it, so it should probably be used instead, deprecating the existing `merge_status`.

marcelstoer:
Fixes #880

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/MergeRequest.java
+++ b/src/main/java/org/gitlab4j/api/models/MergeRequest.java
@@ -36,7 +36,11 @@ public class MergeRequest {
     private Date latestBuildStartedAt;
     private String mergeCommitSha;
     private String squashCommitSha;
+    // see https://docs.gitlab.com/ee/update/deprecations.html#merge_status-api-field
+    // -> deprecated since 15.6, use detailedMergeStatus instead
+    @Deprecated
     private String mergeStatus;
+    private String detailedMergeStatus;
     private Date mergedAt;
     private Participant mergedBy;
     private Boolean mergeWhenPipelineSucceeds;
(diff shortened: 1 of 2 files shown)


src/main/java/org/gitlab4j/api/models/MergeRequest.java
```diff
@@ -36,7 +36,11 @@ public class MergeRequest {
     private Date latestBuildStartedAt;
     private String mergeCommitSha;
     private String squashCommitSha;
+    // see https://docs.gitlab.com/ee/update/deprecations.html#merge_status-api-field
+    // -> deprecated since 15.6, use detailedMergeStatus instead
```
jmini, line 40: I would use a javadoc comment, since compared to a regular comment, this is more helpful inside an IDE. Something like:

```suggestion
    /**
     * @deprecated since 15.6, use {@link #detailedMergeStatus} instead.
     * see https://docs.gitlab.com/ee/update/deprecations.html#merge_status-api-field
     */
```

src/main/java/org/gitlab4j/api/models/MergeRequest.java
```diff
@@ -273,6 +277,7 @@ public void setSquashCommitSha(String squashCommitSha) {
         this.squashCommitSha = squashCommitSha;
     }
 
+    @Deprecated
```
jmini, line 280: Same here, indicate what is the replacement:

```suggestion
    /**
     * @deprecated since 15.6, use {@link #getDetailedMergeStatus()} instead.
     */
    @Deprecated
```

src/main/java/org/gitlab4j/api/models/MergeRequest.java
```diff
@@ -281,6 +286,14 @@ public void setMergeStatus(String mergeStatus) {
         this.mergeStatus = mergeStatus;
```
jmini, line 286: Shouldn't be this setter annotated as well?

marcelstoer: Maybe yes, maybe no. Technically it'd certainly be correct I can't imagine a use case where someone's integration code is going to call the _setter_. You can't actually modify the merge status on GitLab.

jmini:
IMO this change should update the corresponding test: 
https://github.com/gitlab4j/gitlab4j-api/blob/master/src/test/resources/org/gitlab4j/api/merge-request.json

By adding a line like:
```
  "detailed_merge_status": "can_be_merged",
```

See doc: https://docs.gitlab.com/ee/api/merge_requests.html#response

jmini (approved):
Looks good to me

PR #890, opened 2022-12-22
Add headPipeline to MergeRequest

vimsucks:
[https://docs.gitlab.com/ee/api/merge_requests.html#get-single-mr](https://docs.gitlab.com/ee/api/merge_requests.html#get-single-mr)

image

image

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/MergeRequest.java
+++ b/src/main/java/org/gitlab4j/api/models/MergeRequest.java
@@ -43,6 +43,7 @@ public class MergeRequest {
     private String mergeError;
     private Milestone milestone;
     private Pipeline pipeline;
+    private Pipeline headPipeline;
     private Long projectId;
     private String sha;
     private Boolean shouldRemoveSourceBranch;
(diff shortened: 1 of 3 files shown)


src/main/java/org/gitlab4j/api/models/MergeRequest.java
```diff
@@ -329,6 +330,15 @@ public void setPipeline(Pipeline pipeline) {
...
     }
 
+    public Pipeline getHeadPipeline() {
+        return headPipeline;
+    }
+
+    public MergeRequest setHeadPipeline(Pipeline headPipeline) {
+        this.headPipeline = headPipeline;
+        return this;
+    }
```
jmini, line 340: In this project the setters are returning `void`.

When an attribute is used to create or update an entity, there is a builder method `withHeadPipeline(..)` next to the setter.
But in this case the `head_pipeline` attribute is only used during a read, so the builder method makes no sense.

```suggestion
    public void setHeadPipeline(Pipeline headPipeline) {
        this.headPipeline = headPipeline;
    }
```

src/main/java/org/gitlab4j/api/models/Pipeline.java
```diff
@@ -33,6 +34,15 @@ public void setId(Long id) {
...
     }
 
+    public Long getProjectId() {
+        return projectId;
+    }
+
+    public Pipeline setProjectId(Long projectId) {
+        this.projectId = projectId;
+        return this;
+    }
```
jmini, line 44: Same here:

```suggestion
    public void setProjectId(Long projectId) {
        this.projectId = projectId;
    }
```

jmini (changes requested):
I would do minor changes to be consistent with the rest of the code. But for me this change looks good to me.

vimsucks:
> I would do minor changes to be consistent with the rest of the code. But for me this change looks good to me.

thanks! suggestion applied

jmini (approved):

PR #897, opened 2023-01-16
Remove unnesesary runtime charset search

luvarqpp:
Use StandardCharsets.UTF_8 (in java, since java 1.7), instead of "UTF-8" string. This removes necessity to throw a checked exception.

### DIFF
--- a/src/main/java/org/gitlab4j/api/utils/UrlEncoder.java
+++ b/src/main/java/org/gitlab4j/api/utils/UrlEncoder.java
@@ -1,8 +1,7 @@
 package org.gitlab4j.api.utils;
 
 import java.net.URLEncoder;
-
-import org.gitlab4j.api.GitLabApiException;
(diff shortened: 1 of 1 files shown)


jabby (changes requested):
Thanks for your contribution. I will merge it when master will be against Java 17.

jmini:
This change needs `URLEncoder.html#encode(java.lang.String, java.nio.charset.Charset)` which is not available with java 8.

* Java 8 javadoc: https://docs.oracle.com/javase/8/docs/api/java/net/URLEncoder.html
* Java 11 javadoc: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URLEncoder.html#encode(java.lang.String,java.nio.charset.Charset)

jmini (approved):
I have rebased this PR on top of `6.x` (which uses Java 11)

PR #901, opened 2023-01-24
Add id to ProtectedBranch and BranchAccessLevel

jmini:
The protect repository branches API returns `id` attributes in the JSON. See:
https://docs.gitlab.com/ee/api/protected_branches.html#list-protected-branches


jmini:
I think the id were added couple of months ago with:
https://gitlab.com/gitlab-org/gitlab/-/issues/377927

PR #905, opened 2023-02-01
Update EventsApi.java

chenlingmin:
change projectIdOrPath as Object type


jhult:
This is consistent with how this is [everywhere else](https://github.com/gitlab4j/gitlab4j-api/search?q=projectIdOrPath) (an `Object` is used).

[The documentation](https://docs.gitlab.com/ee/api/events.html#list-a-projects-visible-events) also states that `projectIdOrPath` can be an integer/string (The ID or URL-encoded path of the project).

image

So, I think this should be merged.

PR #913, opened 2023-02-15
#912 Add field iteration into IssueFilter

### LINKED ISSUE #912: Add field iteration into IssueFilter
Problem: i need to filter issues by field iteration, but issue filter does not have this field
Solution: add field "iterationTitle" into IssueFilter with type String.

DoubleDPro:
Add iterationTitle in IssueFilter for filtration issues by this field

Fixes #912


jmini (approved):
The `iteration_title` is one of the filter that can be used. Mentioned in the docs here:
https://docs.gitlab.com/ee/api/issues.html#list-project-issues

I am wondering if it makes sense to continue to grow the number of `createIssue(..)` methods. 
In some other area pattern with an object with a builder pattern is used.

But I think it is fine for now.

jabby:
@jmini I think it's fine to add a new `createIssue(..)` method. In a future major update maybe we will remove this kind of overload but it's not in a near future.

PR #928, opened 2023-03-18
Remove three checked GitLabApiException

luvarqpp:
There was not chance of throwing checked exception GitLabApiException  from given three methods. I have removed them to make user code using given methods easier to read.

PS: If throwing GitLabApiException  is wanted effect, I can make 'page' method throw them, instead of other exceptions (unchecked, like 'NoSuchElementException').


jmini (approved):
I don't know the reason why the checked exception were thrown here.

For me it is OK to remove them.

PR #938, opened 2023-04-04
Fix "Results Paging" code example in README.md

turbolocust:
Fixed code example of "Results Paging"


jmini:
Thank you for this contribution!

Issue #894, opened 2023-01-11
Add java9+ and Jakarta EE Support


walnut-tom:
Add java9+ and Jakarta EE Support?

Korde-vaibhav:
When can this release be expected ?

BenderRodrigez:
Just come here while resolving my issues related to Spring 3 migration. Is there a way to use snapshot version?

gedeffe:
Hello,

As we are moving our projects to new version of Spring Boot (3.x), it implies also a move to JEE (jakarta).
Feel free to comment related PR (#943) which has been successfully tested locally with our use of gitlab api ^^

Have a nice day
Gedeffe

jmini:
@gedeffe, @Korde-vaibhav and @walnut-tom 

Please check if the release `6.0.0-rc.1` works for you and give feedback, so that we know if we are on the good path for a `6.0.0` release.

I am closing this issue now.

If you need something more, please reopen this issue with exact details about what is missing. Or open a new one.

PR #942, opened 2023-04-07
Compatibility with Spring Boot 3.x

gedeffe:
In order to use gitlab4j-api with new version of SpringBoot and related dependencies, we had to upgrade several libraries.
It implies also a move from javax.xxx to jakarta.xxx for JEE libraries.


gedeffe:
- Related to Add java9+ and Jakarta EE Support #894 ?

dousp:
[itlab4j-api-5.1.0](https://github.com/gitlab4j/gitlab4j-api/releases/tag/gitlab4j-api-5.1.0)
This means that the above version(5.1.0) is already available in spring Boot3 and JDK 17, right?

Issue #817, opened 2022-03-09
Move to Java 11 or Java 17 for next major iteration of gitlab4j-api


jabby:
As maintainer, I don't want to support Java 8 anymore. 

I want to upgrade to a newer LTS version of Java. I'm not sure if I will jump to latest LTS (Java 17) or if I migrate to Java 11 first. 

If you use gitlab4j-api, feel free to comment this issue.

ludakhris:
So would this mean that you aren’t guaranteeing that it’ll work in Java 8? Or that it won’t run on Java 8 at all? Maybe as a first step go to Java 11, that’s what other libraries are doing.

jabby:
Some updates about updating to next major Java LTS. I tried to summarized what I learned about Java end of life and adoption of Java 17 in the industry. 

Spring 6 baseline will be Java 17+ and Jakarta EE 9. Spring Boot 3 based applications will require a minimum of JDK 17 at runtime.
Quarkus already support Java 17 in JVM mode. Not in native mode. See https://github.com/quarkusio/quarkus/issues/21481 and https://github.com/quarkusio/quarkus/issues/20891 for more informations.

According to JRebel Java Developer Productivity report 2022, `62% of those with knowledge of their upgrade plans reported their intention to move to JDK 17 within the next 12 months.`. 

According to https://endoflife.date/java 
* Java 11 active support will ends at 2023-09-30
* Java 11 security support will ends at 2026-09-30
* Java 17 active support will ends at 2026-09-30
* Java 17 active support will ends at 2031-09-30

As I don't plan to release a new major release of gitlab4j-api before 2023, I think the next major release will require a minimum of JDK 17. 

In the meantime, I will continue to support Java 8 and release new version of gitlab4j-api 5.x

If you have any concern about the future of gitlab4j-api and especially about a JDK 17 requirement, feel free to add a comment here. 

References: 
* https://spring.io/blog/2022/03/28/an-update-on-java-17-adoption
* https://quarkus.io/blog/quarkus-2-6-0-final-released/#java-17-in-generated-projects
* https://github.com/quarkusio/quarkus/issues/21481
* https://spring.io/blog/2021/09/02/a-java-17-and-jakarta-ee-9-baseline-for-spring-framework-6
* https://www.jrebel.com/resources/java-developer-productivity-report-2022

akphi:
[...]

Issue #926, opened 2023-03-17
Start a second branch to work on a version 6 of this library?


jmini:
**Context:**

While there are a lot plans for the future (#817, #924, #925 and other issues with the [Next](https://github.com/gitlab4j/gitlab4j-api/labels/next) label), it seems that we need also to think short term here.

The most wanted feature is to have a version of this library that is compatible with Spring Framework 6.0 (Spring Boot 3) which requires a Jersey client version update and the package change `javax.ws.rs.*` -> `jakarta.ws.rs.*` which is presented in PR https://github.com/gitlab4j/gitlab4j-api/pull/841

---

**Branching model:**

It seems reasonable to me to start a second main branch `jakarta` or `6.x.x` or `master/6.x.x` next to the `master` branch (our current default branch dedicated to `5.1.0`) to put the changes requested by the Spring community to have a version using jakarta.

As commented by @marcelstoer here https://github.com/gitlab4j/gitlab4j-api/pull/841#issuecomment-1348029230 we could have the 2 versions in parallel for a while.

For the branching model, I personally prefer a pattern with forward merge instead of doing cherry-picks.

![git-flow](https://user-images.githubusercontent.com/1222165/225855104-c58dbe88-a8a6-407a-9773-449241158a2f.png)

We have 2 main branches:

* one to work on `5.x.x` (the current version, java 8 &`javax.ws.rs.*` -- could stay the default branch for now)
* one to work on `6.x.x` where we would do have the breaking changes (`jakarta.ws.rs.*` namespace change in particular)

We always merge in the `5.x.x` -> `6.x.x` direction, not in the other way around.
We are sure that all the changes in `5.x.x` are properly merged into `6.x.x`.

💡 I wrote a more [detailed description](https://unblu.github.io/ucascade/user-docs/multi-main.html) about this branching model that we are using in all our projects in my company.

---

**Scope:**

If we want to have something quick, I think it makes sense to keep the `6.x.x` as close as possible to `5.1.x` for now.

So for me this would mean:

* PR #841
* PR #895 (maybe)
* Java 11 as baseline (also not really mandatory)

And do a `6.0.0-rc.1` or a `6.0.0-beta.1` (not sure what is the preferred versioning scheme) really soon, so that we can get feedback from the community that uses Spring.

Other bigger refactoring tasks would be moved to a later version:

* Elimination of the Jersey client #924
* Move to java 17
* Having too much overloaded methods for the same endpoint to handle all the optional parameters
[...]

PR #841, opened 2022-04-26
Migrate from javax to jakarta

yamass:
People are more and more migrating to the newer EE versions (Jakarta). Jetty 11 has been out for more than a year and depends on jakarta. Spring 6 (release data: Q4 2022) will depend on jakarta. https://spring.io/blog/2021/09/02/a-java-17-and-jakarta-ee-9-baseline-for-spring-framework-6

I suggest you introduce two versions of gitlab4j-api: One depending on javax and one on jakarta, maybe on two different branches.

PR #943, opened 2023-04-07
Compatibility with Spring Boot 3.x

gedeffe:
In order to use gitlab4j-api with new version of SpringBoot and related dependencies, we had to upgrade several libraries.
It implies also a move from javax.xxx to jakarta.xxx for JEE libraries.

Relates to existing issues:
- Move to Java 11 or Java 17 for next major iteration of gitlab4j-api #817
- Add java9+ and Jakarta EE Support #894

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<groupId>org.gitlab4j</groupId>
 	<artifactId>gitlab4j-api</artifactId>
 	<packaging>jar</packaging>
-	<version>5.1.0</version>
+	<version>6.0.0</version>
 	<name>GitLab4J-API - GitLab API Java Client</name>
 	<description>GitLab4J-API (gitlab4j-api) provides a full featured Java client library for working with GitLab repositories and servers via the GitLab REST API.</description>
 	<url>https://github.com/gitlab4j/gitlab4j-api</url>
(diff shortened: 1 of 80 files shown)


src/main/java/org/gitlab4j/api/ApplicationSettingsApi.java
```diff
@@ -1,23 +1,20 @@
...
-import com.fasterxml.jackson.databind.JsonNode;
-
 /**
  * This class implements the client side API for the GitLab Application Settings API.
  * See <a href="https://docs.gitlab.com/ee/api/settings.html">Application Settings API at GitLab</a> for more information.
  */
 public class ApplicationSettingsApi extends AbstractApi {
 
-    public ApplicationSettingsApi(GitLabApi gitLabApi) {
+    public ApplicationSettingsApi(final GitLabApi gitLabApi) {
```
jmini, line 17: Why did you make all the method argument explicitly final?

I think the compiler was already treating those as implicitly final

pom.xml
```diff
@@ -297,7 +296,8 @@
 				<configuration>
 					<signature>
 						<groupId>org.codehaus.mojo.signature</groupId>
-						<artifactId>java1${java.level}</artifactId>
+						<artifactId>java18</artifactId>
```
jmini, line 299: Per https://github.com/mojohaus/animal-sniffer/issues/62, keeping animal-sniffer when building with Java 11 does not really make sense.

gedeffe: So, we would be able to remove this configuration ?

gedeffe:
> This is similar to the work that was done in #841 not?
> 
> As I wrote on #926, for me a solution could be to start working on those topic on a separated branch and have a cascade between the `5.x` branch and the `6.x` branch…

If I understand correctly, it would be better to target branch 6.x for these evolutions ?

pom.xml
jmini: This is what I did in https://github.com/gitlab4j/gitlab4j-api/pull/946 (keeping you are co-author of the change)

gedeffe:
@jmini , Thank you for integration in branch 6.x, especially as I made noise with automatic format of code (like adding final keyword).

Do I have to keep this PR alive, or should I edit it to target only 6.x branch or close it ?

jmini:
@ajivemu: Please check if the release `6.0.0-rc.1` works for you and give feedback, so that we know if we are on the good path for a `6.0.0` release.

This version is using:

* minimal version: **Java 11** 
* Jakarta EE components with the `jakarta.*` packages instead of `javax.*`

jmini:
@gedeffe https://github.com/gitlab4j/gitlab4j-api/pull/943#issuecomment-1504727995 :

> Do I have to keep this PR alive, or should I edit it to target only 6.x branch or close it ?

I have open 3 PRs with all your ideas from this PR.

You can review the last open one https://github.com/gitlab4j/gitlab4j-api/pull/950 and tell if something is missing.

For me this PR can be closed. Thank you very much for your work.

gedeffe:
And I forgot to mention also here that version 6.0.0-rc.1 works for our use cases.
Thank you guys !

PR #965, opened 2023-04-25
Make Pager constructor public

jensli:
It is often useful for clients of this library to subclass a `AbstractApi`, or one of its concrete subclasses, to implement custom requests towards the GitLab API.

This is hard right now, because the constructor of `Pager` is package private, so clients cannot access it. This is a bit weird, since `Pager` itself is public. 

This change makes the constructor public.


jmini (approved):
For me the change is OK.

PR #972, opened 2023-05-04
Add Automatic-Module-Name manifest entry

jensli:
This gets rid of warnings on Java 9+, and makes it safer to use this library as a module.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -171,6 +171,9 @@
 				<configuration>
 					<archive>
 						<manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+						<manifestEntries>
+							<Automatic-Module-Name>org.gitlab4j</Automatic-Module-Name>
+						</manifestEntries>
 					</archive>
 				</configuration>
 			</plugin>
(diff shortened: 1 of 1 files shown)


jmini:
The CI error is unrelated with this change. We need to merge https://github.com/gitlab4j/gitlab4j-api/pull/970 or an equivalent solution

pom.xml
```diff
@@ -171,6 +171,9 @@
 				<configuration>
 					<archive>
 						<manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+						<manifestEntries>
+							<Automatic-Module-Name>org.gitlab4j</Automatic-Module-Name>
```
jmini, line 175: Where is the best practice definition for automatic-module-names?

since our *groupId* is: `org.gitlab4j`
and our *artifactId* is: `gitlab4j-api`

I am wondering if it wouldn't be better to name this module `org.gitlab4j.api`…

jensli: I don't know. Your suggestion `org.gitlab4j.api` absolutely sounds reasonable.

jmini: @jabby do you have an opinion?

jabby: I think `org.gitlab4j.api` is a good idea.

jensli: > I think `org.gitlab4j.api` is a good idea.

Updated to `org.gitlab4j.api`.

jmini (approved):

Issue #925, opened 2023-03-17
Move the test infrastructure to Testcontainers


jmini:
Currently some integration are running against a GitLab started in docker, controlled by those properties in the POM:

https://github.com/gitlab4j/gitlab4j-api/blob/b2942f39fc516c584cdce1567bf79638e222e834/pom.xml#L69-L72

* The GitLab version version is old.
* Tests rely on maven starting the container using the `io.fabric8:docker-maven-plugin` plugin.
* It is not easy to skip the integration tests on a computer where docker is not installed.

The [Testcontainers](https://www.testcontainers.org/) is a popular project where integration tests declare their dependencies directly in the code (next to the tests).

---

It would be also handy to be able to test against multiple GitLab versions, for example the one listed in the version selector on https://docs.gitlab.com/.

<img width="249" alt="GitLab version selector in the docs" src="https://user-images.githubusercontent.com/1222165/225846676-eb73df00-dc20-4d0a-9eeb-15e90a491f1a.png">

For example as of today this would be:

* `15.9`
* `14.10`
* `13.10`

---

_This is open for suggestions_

jmini:
I am working on an other approach for our integration tests. See: https://github.com/gitlab4j/gitlab4j-api/issues/1165

PR #973, opened 2023-05-04
added response headers to GitLabApiException

Walti91:


jmini:
The CI error is unrelated with this change. We need to merge https://github.com/gitlab4j/gitlab4j-api/pull/970 or an equivalent solution

jmini (approved):
The change looks good to me.

I have rebased the change on top of `main` and I have fixed the Unit Test that was failing (probably because we are testing against a very old GitLab instance (see https://github.com/gitlab4j/gitlab4j-api/issues/925 --> our plan to improve on this topic).

jmini:
@Walti91 thank you for this contribution 🎉 

It was merged to `main` and then to `6.x` and will be available with the next release.

Issue #1014, opened 2023-08-04
Add missing support for listing LDAP group links


profhenry:
With https://github.com/gitlab4j/gitlab4j-api/issues/223 support for LDAP group links were added.

Unfortunately the method for listing the LDAP group links was not implemented :-/.

See https://docs.gitlab.com/ee/api/groups.html#list-ldap-group-links

PR #1015, opened 2023-08-04
Added support for listing LDAP group links

profhenry:
Fixes https://github.com/gitlab4j/gitlab4j-api/issues/1014

See https://docs.gitlab.com/ee/api/groups.html#list-ldap-group-links

---

About the `LdapGroupLink` model:

> The GitLab REST API documentation (https://docs.gitlab.com/ee/api/groups.html#list-ldap-group-links)
> does not specify the attributes which are returned for a LDAP group link :-/.
> This are the attributes returned by our GitLab instance (`15.10.2-ee`).
> These seem fine because they match with the attributes which you can pass when adding a LDAP group link :-).

### DIFF
--- a/src/main/java/org/gitlab4j/api/models/LdapGroupLink.java
+++ b/src/main/java/org/gitlab4j/api/models/LdapGroupLink.java
@@ -5,11 +5,6 @@ import org.gitlab4j.api.utils.JacksonJson;
 
 public class LdapGroupLink {
 
-    // The GitLab REST API documentation (https://docs.gitlab.com/ee/api/groups.html#list-ldap-group-links)
-    // does not specify the attributes which are returned for a LDAP group link :-/.
-    // This are the attributes returned by our GitLab instance (15.10.2-ee).
-    // These seem fine because they match with the attributes which you can pass when adding a LDAP group link :-).
-
     private String cn;
 
     private AccessLevel groupAccess;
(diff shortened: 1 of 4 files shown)


jmini (changes requested):
We should have an example JSON for `LdapGroupLink.java` tested like the other model class.

Otherwise the change looks good to me.

src/main/java/org/gitlab4j/api/models/LdapGroupLink.java
```diff
@@ -0,0 +1,57 @@
...
+package org.gitlab4j.api.models;
+
+import org.gitlab4j.api.utils.JacksonJson;
+
+public class LdapGroupLink {
+
+    // The GitLab REST API documentation (https://docs.gitlab.com/ee/api/groups.html#list-ldap-group-links)
+    // does not specify the attributes which are returned for a LDAP group link :-/.
+    // This are the attributes returned by our GitLab instance (15.10.2-ee).
+    // These seem fine because they match with the attributes which you can pass when adding a LDAP group link :-).
```
jmini, line 11: I appreciate this comment, it is very valuable.

But I think we can remove it from this code base.

jmini (approved):
Looks good to me.

I hope the `ldap-group-link.json` I wrote is realistic enough (but I do not have a GitLab with LDAP to verify this)

jmini:
merged! 

@profhenry Thank you very much for your contribution and for the feedback here.

PR #963, opened 2023-04-22
UserApi implement createPersonalAccessToken

jmini:
* adds the corresponding method
* adds missing scopes
* factors out common logic

Fixes #653

---

This is a rework of #686 from @maltem-za (his PR can't be edited):

* Rebased on top of the main branch
* Renaming `ImpersonationToken` into `PersonalAccessToken` is breaking from a java point of view and because of the nested enum `ImpersonationToken.Scope` it is really difficult to change in a backward compatible way.
    * This PR is keeping the old name
    * Once this is merged, on the branch `6.x` the class can be renamed.

PR #1018, opened 2023-08-16
Add Project Access Tokens

### LINKED ISSUE #693: Add support for project access token CRUD
Gitlab has now added the ability to CRUD project access tokens:
https://docs.gitlab.com/13.10/ee/api/resource_access_tokens.html

The ability to create and interact with project access tokens would be a powerful addition to the current features exposed by the sdk

bensofficial:
For further development within [the Artemis project](https://github.com/ls1intum/Artemis) we need project access tokens. Therefore, we implemented the corresponding feature.

This PR adds the methods to interact with the [project access tokens API provided by GitLab](https://docs.gitlab.com/ee/api/project_access_tokens.html).

I have tested the newly introduced methods with GitLab version 13.10.

If you have questions, please let us know.

Closes #693


src/main/java/org/gitlab4j/api/ProjectApi.java
```diff
@@ -3901,4 +3877,92 @@ public RemoteMirror updateRemoteMirror(Object projectIdOrPath, Long mirrorId, Bo
...
+            .withParam("name", name, true)
+            .withParam("expires_at", ISO8601.dateOnly(expiresAt), true)
+            .withParam("scopes", scopes, true)
+            .withParam("access_level", (Object) null, false);
+        Response response = post(Response.Status.CREATED, formData,
+            "projects", getProjectIdOrPath(projectIdOrPath), "access_tokens");
+        return (response.readEntity(ProjectAccessToken.class));
+    }
+
+    // TODO
```
jmini, line 3951: That is this TODO? Just the missing javadoc

src/main/java/org/gitlab4j/api/ProjectApi.java
```diff
@@ -38,31 +38,7 @@
...
-import org.gitlab4j.api.models.ProjectGroupsFilter;
-import org.gitlab4j.api.models.ProjectGroup;
-import org.gitlab4j.api.models.ProjectHook;
-import org.gitlab4j.api.models.ProjectUser;
-import org.gitlab4j.api.models.PushRules;
-import org.gitlab4j.api.models.RemoteMirror;
-import org.gitlab4j.api.models.Snippet;
-import org.gitlab4j.api.models.Variable;
-import org.gitlab4j.api.models.Visibility;
+import org.gitlab4j.api.models.*;
```
jmini, line 41: Not sure this change is really useful…

It might toogle back with the next contribution, which create necessary diffs.

src/main/java/org/gitlab4j/api/models/ProjectAccessToken.java
```diff
@@ -0,0 +1,118 @@
...
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.gitlab4j.api.Constants;
+import org.gitlab4j.api.utils.JacksonJson;
+
+import java.util.Date;
+import java.util.List;
+
+public class ProjectAccessToken {
```
jmini, line 11: There should be a json example and a corresponding test in `TestGitLabApiBeans`

bensofficial:
Thank you for your feedback 🙏 . I implemented your suggestions.

bensofficial:
Regarding the failing test case: Some of the API endpoints have been added in GitLab release 14.10 and 16.0 (https://gitlab.com/gitlab-org/gitlab/-/merge_requests/82714 or https://gitlab.com/gitlab-org/gitlab/-/merge_requests/120436). As the tests currently run against GitLab 13.9 the newly added test cases will fail for an obvious reason.
Maybe I should disable the tests until the version of the GitLab instance is updated in the `pom.xml`. What do you think @jmini?

jmini:
> As the tests currently run against GitLab 13.9 the newly added test cases will fail for an obvious reason.
> Maybe I should disable the tests until the version of the GitLab instance is updated in the pom.xml. 

Yes I think we already did something like that in https://github.com/gitlab4j/gitlab4j-api/pull/963/ just adding a `NOTE`.

There is an idea to switch the GitLab instance that used in the integration tests #925, but I don't know if there are any progress.
Once we have a more recent version, we can enable back those tests.

bensofficial:
Great. I will add it tomorrow, and then the PR is good to go 😄

bensofficial:
@jmini I have now commented out the affected lines. Everything works now 👍

jmini (approved):

jmini:
Merged, thank you very much for this contribution! 🎉

PR #1021, opened 2023-08-22
Add namespace to user

bensofficial:
The problem arising is that the namespace id is not always equal to the user id.

Example:

1. Create user: User Id 1, Namespace Id 1
2. Create Group: Namespace Id 2
3. Create user: User Id 2, Namespace Id 3

This PR adds the attribute to the user so that the information is available in a user object.


jmini (approved):

jmini:
Merged, thank you very much for this contribution! 🎉

PR #1029, opened 2023-09-08
Add the "internal" flag to NotesApi.createIssueNote()

hannibal218bc:
- add a new `createIssueNote()` method allowing to pass the `internal` flag.
- add the `internal` flag to the `Note` model

### DIFF
--- a/src/main/java/org/gitlab4j/api/NotesApi.java
+++ b/src/main/java/org/gitlab4j/api/NotesApi.java
@@ -152,7 +152,7 @@ public class NotesApi extends AbstractApi {
      * @throws GitLabApiException if any exception occurs
      */
     public Note createIssueNote(Object projectIdOrPath, Long issueIid, String body) throws GitLabApiException {
-        return (createIssueNote(projectIdOrPath, issueIid, body, null));
+        return (createIssueNote(projectIdOrPath, issueIid, body, null, null));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/gitlab4j/api/NotesApi.java
```diff
@@ -166,10 +166,28 @@ public Note createIssueNote(Object projectIdOrPath, Long issueIid, String body)
...
+    public Note createIssueNote(Object projectIdOrPath, Long issueIid, String body, Date createdAt, boolean internal) throws GitLabApiException {
 
         GitLabApiForm formData = new GitLabApiForm()
                 .withParam("body", body, true)
-                .withParam("created_at", createdAt);
+                .withParam("created_at", createdAt)
+        ;
+        if (internal) {
+        	formData.withParam("internal", "true");
+        }
```
jmini, line 190: No need to check for the internal value. If you use `Boolean` the `withParam` adds it only if not null.

```suggestion
                .withParam("created_at", createdAt)
                .withParam("internal", internal);
```

jmini (changes requested):
Please also add the `internal` flag to the example file:
https://github.com/gitlab4j/gitlab4j-api/blob/main/src/test/resources/org/gitlab4j/api/note.json
--> this should be as close as possible to a real object returned by GitLab.

This way this test make sure our model matches with the JSON example:

https://github.com/gitlab4j/gitlab4j-api/blob/ff037bd36355a7391d3ba7a7c417410d3cbf3612/src/test/java/org/gitlab4j/api/TestGitLabApiBeans.java#L681-L685

... and your addition is covered checked by a test.

jmini:
@hannibal218bc thank you for this PR. Small changes are requested but the change looks almost good to me.

hannibal218bc:
> @hannibal218bc thank you for this PR. Small changes are requested but the change looks almost good to me.

@jmini thank you for the feedback. I'll submit the changes as soon I as time permits.

src/main/java/org/gitlab4j/api/NotesApi.java
```diff
@@ -152,7 +152,7 @@ public Note getIssueNote(Object projectIdOrPath, Long issueIid, Long noteId) thr
      * @throws GitLabApiException if any exception occurs
      */
     public Note createIssueNote(Object projectIdOrPath, Long issueIid, String body) throws GitLabApiException {
-        return (createIssueNote(projectIdOrPath, issueIid, body, null));
+        return (createIssueNote(projectIdOrPath, issueIid, body, null, false));
```
jmini, line 155:
```suggestion
        return (createIssueNote(projectIdOrPath, issueIid, body, null, null));
```

hannibal218bc:
@jmini thank you for the review, I've updated the MR with the requested changes.

> Please also add the `internal` flag to the example file: https://github.com/gitlab4j/gitlab4j-api/blob/main/src/test/resources/org/gitlab4j/api/note.json --> this should be as close as possible to a real object returned by GitLab.

I cross-checked the file against a file retrieved from a 16.4 enterprise instance, and there are a few more differences - added `attachments`, `confidential`,  but `author.created` is missing. And a new field `type` that isn't even documented at Gitlab...

I guess the JSON shall only contain properties that are actually supported by Gitlab4J?

jmini:
> I guess the JSON shall only contain properties that are actually supported by Gitlab4J?

Sadly this is the long work of keeping our model aligned with the JSON returned by GitLab. Step by step we are getting closer.

I agree that the field you mention are out of scope for this pull request.

> And a new field `type` that isn't even documented at Gitlab

This is crazy. They have also issue keeping up-to-date their doc with their code.

jmini (approved):
It looks good to me. I will do some manual testing before merging.

PR #1036, opened 2023-10-01
#982 unable to find path for getArchive api

### LINKED ISSUE #982: unable to find path for getArchive api
for api

https://docs.gitlab.com/ee/api/repositories.html#get-file-archive

i can't see how to specify the path parameter (that should fetch just part of repository)

can you give me an hint please? :)

sing 6.0.0 rc1

DoubleDPro:
Added "path" parameter to methods

Fixes #982

### DIFF
--- a/src/main/java/org/gitlab4j/api/RepositoryApi.java
+++ b/src/main/java/org/gitlab4j/api/RepositoryApi.java
@@ -22,6 +22,7 @@ import org.gitlab4j.api.models.ChangelogPayload;
 import org.gitlab4j.api.models.Commit;
 import org.gitlab4j.api.models.CompareResults;
 import org.gitlab4j.api.models.Contributor;
+import org.gitlab4j.api.models.RepositoryArchiveParams;
 import org.gitlab4j.api.models.TreeItem;
 import org.gitlab4j.api.utils.FileUtils;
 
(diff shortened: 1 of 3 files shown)


jmini:
Why did you try to `ByPathAndSha` to each of the methods?  Was it mandatory because of a naming clash?

DoubleDPro:
> Was it mandatory because of a naming clash?

Yes. There is a naming clash with deprecated method so I decided to rename new methods. If you have a better solution, I will correct my PR.

jmini:
Sorry for the delay.

I think it would be cleaner to add a `RepositoryArchiveParams` providing all the possible attributes with a builder pattern `withSha(String)`, `withPath(String)`, `withFormat(ArchiveFormat)` and so on.

And then we would have a method:

* `getRepositoryArchive(Object projectIdOrPath, RepositoryArchiveParams params)` returning an `InputStream`
* `getRepositoryArchive(Object projectIdOrPath, RepositoryArchiveParams params, File directory)` returning a `File`

The existing methods can delegate to the new one.

This way we have the maximum flexibility of providing the parameters we would like. Methods overloading is no longer an issue.

This is a pattern we already have at multiple places.

@DoubleDPro what do you think?

DoubleDPro:
@jmini it looks good to me. I will correct it.

DoubleDPro:
@jmini I corrected my PR. Can you check it?

jmini (approved):
Looks good to me.

I will perform some testing and then merge

Issue #1042, opened 2023-10-12
Design alias reopenIssue for updateIssue(..., Constants.StateEvent.REOPEN, ...)


neilwangweili:
I see an api for closing issues.
~~~java
gitlabApi.getIssueApi().closeIssue(1L, 2L);
~~~

But it doesn't look like there's an api to reopen issues fast. I need to do this to reopen issue:
~~~java
api.getIssuesApi().updateIssue(..., Constants.StateEvent.REOPEN, ...);
~~~

But I want to do like this:
~~~java
api.getIssuesApi().reopenIssue(1L, 2L);
~~~

## Solutions
~~~java
public Issue reopenIssue(Object projectIdOrPath, Long issueIid) throws GitLabApiException {

        if (issueIid == null) {
            throw new RuntimeException("issue IID cannot be null");
        }

        GitLabApiForm formData = new GitLabApiForm().withParam("state_event", StateEvent.REOPEN);
        Response response = put(Response.Status.OK, formData.asMap(), "projects", getProjectIdOrPath(projectIdOrPath), "issues", issueIid);
        return (response.readEntity(Issue.class));
}
~~~
Add it into https://github.com/gitlab4j/gitlab4j-api/blob/main/src/main/java/org/gitlab4j/api/IssuesApi.java

neilwangweili:
Related PR: #1043

PR #1043, opened 2023-10-12
feat: add reopenIssue api in IssuesApi #1042

neilwangweili:
Fixes: #1042


jmini:
I agree that this addition is consistent with the other methods provided in the API.

With other endpoints we have only one update method which accepts a `EntityUpdateParam` or directly the `Entity` and call the update method, but this is not the case for the issue update endpoint.

jmini:

PR #1051, opened 2023-11-08
feat: Add allow_force_push in ProtectedBranch

egore:
See https://docs.gitlab.com/ee/api/protected_branches.html

This also applies to the 6.x branch


jmini (changes requested):
Looks good to me.

Please also update the file `src/test/resources/org/gitlab4j/api/protected-branch.json` so that your change is also covered by unit tests.

Those example files should be as close as possible to the real response sent by GitLab (it can also be the example taken from the documentation page) just we do not support `null` values in the test, so you need to have one explicit value (with the appropriate type) for each attribute (even if in reality 2 attributes can not be set at the same time).

jmini (approved):
I have updated the example file

PR #1072, opened 2023-12-19
added new api call for forced push checks #1071

### LINKED ISSUE #1071: ProtectedBranch contains allowForcePush, but cannot persist it
With #1051 there was the forced_push flag added. So far so good, but I did not find any way how I can persist this feature?

I would have expected it here:
https://github.com/gitlab4j/gitlab4j-api/blob/f7441c791ad2ceb9151dc0c1410adc85d500dcf8/src/main/java/org/gitlab4j/api/ProtectedBranchesApi.java#L122
as one of the protectBranch* methods, but cannot find there

dpalic:
Here a change to expose the forced commit check

Gitlab documentation: https://docs.gitlab.com/ee/api/protected_branches.html

Fixes #1071

### DIFF
--- a/src/main/java/org/gitlab4j/api/ProtectedBranchesApi.java
+++ b/src/main/java/org/gitlab4j/api/ProtectedBranchesApi.java
@@ -136,7 +136,7 @@ public class ProtectedBranchesApi extends AbstractApi {
      * @throws GitLabApiException if any exception occurs
      */
     public ProtectedBranch protectBranch(Object projectIdOrPath, String branchName, AccessLevel pushAccessLevel, AccessLevel mergeAccessLevel) throws GitLabApiException {
-        return (protectBranch(projectIdOrPath, branchName, pushAccessLevel, mergeAccessLevel, null, null));
+        return (protectBranch(projectIdOrPath, branchName, pushAccessLevel, mergeAccessLevel, null, null, null));
     }
 
     /**
@@ -155,13 +155,14 @@ public class ProtectedBranchesApi extends AbstractApi {
      */
     public ProtectedBranch protectBranch(Object projectIdOrPath, String branchName,
             AccessLevel pushAccessLevel, AccessLevel mergeAccessLevel, AccessLevel unprotectAccessLevel,
-            Boolean codeOwnerApprovalRequired) throws GitLabApiException {
+            Boolean codeOwnerApprovalRequired, Boolean allowForcedPush) throws GitLabApiException {
         Form formData = new GitLabApiForm()
                 .withParam("name", branchName, true)
                 .withParam("push_access_level", pushAccessLevel)
                 .withParam("merge_access_level", mergeAccessLevel)
                 .withParam("unprotect_access_level", unprotectAccessLevel)
-                .withParam("code_owner_approval_required", codeOwnerApprovalRequired);
+                .withParam("code_owner_approval_required", codeOwnerApprovalRequired)
+                .withParam("allow_force_push", allowForcedPush);
         Response response = post(Response.Status.CREATED, formData.asMap(),
                 "projects", getProjectIdOrPath(projectIdOrPath), "protected_branches");
         return (response.readEntity(ProtectedBranch.class));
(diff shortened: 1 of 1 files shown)


dpalic:
with the new 6.0 release the api change may be okay, since everyone expects to do minor changes on major change

src/main/java/org/gitlab4j/api/ProtectedBranchesApi.java
```diff
@@ -228,6 +230,7 @@ public ProtectedBranch protectBranch(Object projectIdOrPath, String branchName,
             allowedToMerge.getForm(formData, "allowed_to_merge");
         if (allowedToUnprotect != null)
             allowedToUnprotect.getForm(formData, "allowed_to_unprotect");
+        //FIXME: cannot test, having CE only to test
```
jmini, line 233: I don't think we need this FIXME line in the code.

```suggestion
```

If you know how we could have access to a Testing-Ultimate version, that would be great. I have this on my list, but I would need to ask somebody at GitLab.

src/main/java/org/gitlab4j/api/ProtectedBranchesApi.java
```diff
@@ -155,13 +155,14 @@ public ProtectedBranch protectBranch(Object projectIdOrPath, String branchName,
      */
     public ProtectedBranch protectBranch(Object projectIdOrPath, String branchName,
             AccessLevel pushAccessLevel, AccessLevel mergeAccessLevel, AccessLevel unprotectAccessLevel,
-            Boolean codeOwnerApprovalRequired) throws GitLabApiException {
+            Boolean codeOwnerApprovalRequired, Boolean allowForcedPush) throws GitLabApiException {
```
jmini, line 158: Usually when we add a new parameter we also add a method with the old signature (using java method overloading) to preserve backward compatibility.

src/main/java/org/gitlab4j/api/ProtectedBranchesApi.java
```diff
@@ -185,14 +186,15 @@ public ProtectedBranch protectBranch(Object projectIdOrPath, String branchName,
      */
     public ProtectedBranch protectBranch(Object projectIdOrPath, String branchName,
             Integer allowedToPushUserId, Integer allowedToMergeUserId, Integer allowedToUnprotectUserId,
-            Boolean codeOwnerApprovalRequired) throws GitLabApiException {
+            Boolean codeOwnerApprovalRequired, Boolean allowForcedPush) throws GitLabApiException {
```
jmini, line 189: Usually when we add a new parameter we also add a method with the old signature (using java method overloading) to preserve backward compatibility.

jmini (changes requested):
The change looks good to me, but the changes described in the comments are needed

dpalic:
changed the reported issues, would be great if you could check and merge into master for next major release

src/main/java/org/gitlab4j/api/ProtectedBranchesApi.java
```diff
@@ -150,23 +151,43 @@ public ProtectedBranch protectBranch(Object projectIdOrPath, String branchName,
...
-            Boolean codeOwnerApprovalRequired) throws GitLabApiException {
-        Form formData = new GitLabApiForm()
+            Boolean codeOwnerApprovalRequired, Boolean allowForcedPush) throws GitLabApiException {
+        GitLabApiForm formData = new GitLabApiForm()
                 .withParam("name", branchName, true)
                 .withParam("push_access_level", pushAccessLevel)
                 .withParam("merge_access_level", mergeAccessLevel)
                 .withParam("unprotect_access_level", unprotectAccessLevel)
                 .withParam("code_owner_approval_required", codeOwnerApprovalRequired);
+        if (Objects.nonNull(allowForcedPush)) {
```
jmini, line 167: The `if` here is not necessary. The method `.withParam(..)` is already taking care of that.

jmini (approved):
I made some additional changes.
I will perform some manual tests and merge

src/main/java/org/gitlab4j/api/ProtectedBranchesApi.java
jmini: Fixed with 7ad47ca0769bc88867208a1e2756d5e8d266e5fa

PR #1074, opened 2023-12-20
Models implements serializable interface to support native image(GraalVM)

nilchina:
Hi guys! When building a native image, the classes under the models package need to be registered in the form of serialization, so they need to implement the serialization interface.

### DIFF
--- a/src/main/java/org/gitlab4j/api/GitLabApiException.java
+++ b/src/main/java/org/gitlab4j/api/GitLabApiException.java
@@ -22,8 +22,8 @@ import com.fasterxml.jackson.databind.JsonNode;
  * with a GitLab API endpoint.
  */
 public class GitLabApiException extends Exception {
-
     private static final long serialVersionUID = 1L;
+
     private StatusType statusInfo;
     private int httpStatus;
     private String message;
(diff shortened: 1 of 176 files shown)


nilchina:
@Configuration
@ImportRuntimeHints(MyRuntimeHints.class)
public class MyRuntimeHints implements RuntimeHintsRegistrar {
    @Override
    public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
        hints.serialization().registerType(org.gitlab4j.api.models.Project.class);
        hints.serialization().registerType(org.gitlab4j.api.models.Namespace.class);
        //......
    }
}

jmini:
This is fine for me. I think this PR could target `main` since I am merging all the changes on `main` also to `6.x` (see https://github.com/gitlab4j/gitlab4j-api/issues/926#issue-1628938961)

We have a test class `TestGitLabApiBeans` to test all the models to JSON serialization and deserialization.
I make sure it stays up to date with each PR.

Probably we could add somewhere a test to make sure each class and its sub-class are implementing `Serializable`. Otherwise we can often miss one case.

jmini (approved):

PR #1075, opened 2023-12-20
Add debugging flag

jmini:
Add a flag to log the HTTP trafic


jmini:
This change was reverted. The method already exists:
```
GitLabApi(..., ....)
            .withRequestResponseLogging(Level.INFO);
```

PR #1079, opened 2024-01-21
Add NamespaceApi.getNamespace(idOrPath) using /namespaces/:id

ddoleye:
Just like `GroupApi.getGroup(idOrPath)` and `ProjectApi.getProject(idOrPath)`, 
`getNamespace(idOrPath)` method has been added to `NamespaceApi` as well.

See: https://docs.gitlab.com/ee/api/namespaces.html#get-namespace-by-id


jmini:
This looks like a good addition, but according to our branching model it should target `main`. See https://github.com/gitlab4j/gitlab4j-api/issues/926#issue-1628938961

I will then cascade the change to `6.x`.

jmini:
I have changed the target branch of this PR.

```
git rebase --onto origin/main origin/6.x
git push --force
```

Be aware that your local branch `6.x` has diverged with the one on the GitHub.
If you need to do one more changes on this branch:
I recommend to use `git fetch` and `git reset --hard origin/6.x` before doing any local modifications instead of `git pull`.

Issue #1107, opened 2024-03-27
Add name attribute to project import and export api


schott89:
The name of the imported Project currently results of the given path. So it is impossible to set project names with whitespaces.

POST /projects/import offers the "name" parameter so the project could be named directly.

schott89:
closed with #1108

PR #1108, opened 2024-03-27
Add import function with name parameter

schott89:
contributes to #1107


jmini:
Thank you for this contribution.

Wouldn't it make sense to extract the name from the `overrideParams` parameter. The Project type has a `getName()` that could be used.

schott89:
sounds good

Issue #634, opened 2020-11-13
Unprotect branches with delete


axel1200:
https://github.com/gitlab4j/gitlab4j-api/blob/1287b2363927d6bf5e536a0373cbce28b212a7a6/src/main/java/org/gitlab4j/api/RepositoryApi.java#L246
I was told that this is is supposed to happen via a DELETE instead of PUT https://gitlab.com/gitlab-org/gitlab/-/issues/280805

PR #1112, opened 2024-04-16
Unprotect branches with delete #634 Fixed

AlexHerting:


jmini:
This requires more tests.

I did not find the endpoint in the documentation, but I found a test case here:
https://gitlab.com/gitlab-org/gitlab/-/blob/87385d7e0203c37dc97d4f9845cfe9860c33d2f3/spec/requests/api/branches_spec.rb#L629

Which is using `PUT` and not `DELETE`

jmini:
The documentation says:

https://docs.gitlab.com/ee/api/protected_branches.html#unprotect-repository-branches

```
DELETE /projects/:id/protected_branches/:name
```

PR #1113, opened 2024-04-23
Fix additional long ids

Sineaggi:
Found a few cases where deserialization was throwing errors due to ints being out of out-of-range.

Follow-up to https://github.com/gitlab4j/gitlab4j-api/pull/811

### DIFF
--- a/src/main/java/org/gitlab4j/api/PipelineApi.java
+++ b/src/main/java/org/gitlab4j/api/PipelineApi.java
@@ -10,6 +10,7 @@ import javax.ws.rs.core.GenericType;
 import javax.ws.rs.core.Response;
 
 import org.gitlab4j.api.models.Bridge;
+import org.gitlab4j.api.models.Job;
 import org.gitlab4j.api.models.Pipeline;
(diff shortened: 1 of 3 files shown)


jmini:
The change `int` -> `Long` in the models is a good idea, but it is breaking on the `5.x` stream. Would it be OK to do this change only on `6.x`?

gwidion:
> The change `int` -> `Long` in the models is a good idea, but it is breaking on the `5.x` stream. Would it be OK to do this change only on `6.x`?

If by "breaking" you mean fixing something that had no chance of working, then yes. 🙂

jmini:
Using `int` for the id will work only in on-prem setup (at the beginning, until the system reach the limit of `int` and then start to overflow).

Probably switching to Long is not hurting readers of this value too much.

Sineaggi:
The test TestAvatarUpload keeps 502ing

jmini:
I finally got a green build and I merged your change to `main` and `6.x`. Thank you 🎉 .

Issue #1121, opened 2024-05-07
Missing feature of creating a note-webhook


06luki06:
The ProjectAPI.addHook() method misses the functionality of adding a note-hook using this method.

I tried to create a pull request for this - but unfortunately, I do not have write-permissions for this repo.

jmini:
PR is welcomed.

> I tried to create a pull request for this - but unfortunately, I do not have write-permissions for this repo.

You need to create your own fork of this repo.

See: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork

06luki06:
@jmini ah, alright - thank you! 🙂

PR #1123, opened 2024-05-15
Update ProjectApi.java

06luki06:
Add the possibility of adding noteEvents via the addHook().

Fixes: #1121

### DIFF
--- a/src/main/java/org/gitlab4j/api/ProjectApi.java
+++ b/src/main/java/org/gitlab4j/api/ProjectApi.java
@@ -2211,7 +2211,7 @@ public class ProjectApi extends AbstractApi implements Constants {
      * @throws GitLabApiException if any exception occurs
      */
     public ProjectHook addHook(Object projectIdOrPath, String url, ProjectHook enabledHooks,
-            boolean enableSslVerification, String secretToken) throws GitLabApiException {
+            Boolean enableSslVerification, String secretToken) throws GitLabApiException {
 
         GitLabApiForm formData = new GitLabApiForm()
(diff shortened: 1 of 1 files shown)


jmini:
I have made a change to preserve the old `addHook(..)` method to avoid breaking changes.

I also think that `org.gitlab4j.api.ProjectApi.addHook(Object, String, ProjectHook, Boolean, String)` is the method that is maintained and that should be used. So your method is now delegating to this method.

MarkEWaite:
I believe that the API change of the fourth argument `boolean` (primitive type, nulls not allowed) to `Boolean` (object, nulls allowed) is a breaking change.  The change in [the diff](https://github.com/gitlab4j/gitlab4j-api/pull/1123/files#diff-9e127234e0f2bd8027b49f6e23304c0dd323640d2a9847a027ebae9d882fc7c7R2214) went from this:

```
    public ProjectHook addHook(Object projectIdOrPath, String url, ProjectHook enabledHooks,
            boolean enableSslVerification, String secretToken) throws GitLabApiException {
```

to this:

```
    public ProjectHook addHook(Object projectIdOrPath, String url, ProjectHook enabledHooks,
            Boolean enableSslVerification, String secretToken) throws GitLabApiException {
```

The Jenkins issue tracker received two reports of method not found exceptions in issue [JENKINS-73672](https://issues.jenkins.io/browse/JENKINS-73672).  

Is there a critical need for the change of type in the argument?

If not, could that change be reverted and a new release provided so that API compatibility is retained?

PR #1125, opened 2024-05-16
Expose repository_object_format via Project API

slonopotamus:
GitLab 16.9 introduced an option to create repositories with SHA256 instead of SHA1.

This commit exposes object format used by particular repository via Project API.

See https://gitlab.com/gitlab-org/gitlab/-/issues/419887


src/main/java/org/gitlab4j/api/models/Project.java
jmini, line 1: Can you please also update the json file `/gitlab4j-api/src/test/resources/org/gitlab4j/api/project.json` that corresponds to this model and that is used in the `TestGitLabApiBeans` test.

src/test/java/org/gitlab4j/api/TestProjectApi.java
```diff
@@ -224,6 +224,7 @@ public void testCreate() throws GitLabApiException {
                 .withMergeRequestsEnabled(true)
                 .withWikiEnabled(true)
                 .withSnippetsEnabled(true)
+                .withRepositoryObjectFormat("sha1")
```
jmini, line 227: If I am not mistaken this test is using the GitLab installation started in Docker while the integration tests are running.

This version is very old and this test will fail.

slonopotamus: Hmm... Is there any reference test that I could use for this new field?

slonopotamus:
@jmini is it ok that repositoryObjectFormat is a string? Or it would be better if it was a enum, like project vsibility?

src/test/java/org/gitlab4j/api/TestProjectApi.java
jmini: The pipeline in GitHub action is failing as expected:
```
Error:  Failures: 
Error:    TestProjectApi.testCreate:239 expected: <sha1> but was: <null>
[INFO] 
Error:  Tests run: 308, Failures: 1, Errors: 0, Skipped: 14
```

jmini:
> is it ok that repositoryObjectFormat is a string? Or it would be better if it was a enum, like project vsibility?

I am always not really sure about the String vs enum… In the docs they just mention `sha1` being the default value but they do not say it is sha1 or sha256. So I think String is OK.

src/test/java/org/gitlab4j/api/TestProjectApi.java
jmini: Some tests are using Mockito but this feels also wrong.
I started to experiment to use testcontainers to run a newer GitLab version (there is some work in a PR).

But I am wondering if we should not just Mock the GitLab server in order to run the unit-test quickly.

I would say that for now `TestProjectApi.testCreate()` remains unchanged and we merge your change.

slonopotamus:
> In the docs they just mention `sha1` being the default value but they do not say it is `sha1` or `sha256`

Yes, `sha1` is default. The repository can either be created with `sha1` or `sha256`. It cannot be changed after repository ws created. And it is very unlikely any new values will emerge in nearest 10 years or so.

> So I think String is OK.

Ok.

slonopotamus:
Converting to draft until I fix test issues.

PR #1128, opened 2024-05-28
Update project model

georgeorfanidi:
Add following attributes:
* `merge_requests_template`
* `issues_template`
https://docs.gitlab.com/ee/api/projects.html#templates-for-issues-and-merge-requests


src/main/java/org/gitlab4j/api/models/Project.java
jmini, line 1: can you please update the `project.json` file used in the unit test `TestGitLabApiBeans`?

PR #1132, opened 2024-06-18
improve ProjectHook setting

titilami:
add description for webhook


src/main/java/org/gitlab4j/api/models/ProjectHook.java
jmini, line 1: Please add a `description` attribute to the JSON file [`src/test/resources/org/gitlab4j/api/hook.json`](https://github.com/gitlab4j/gitlab4j-api/blob/main/src/test/resources/org/gitlab4j/api/hook.json) used in tests.

jmini (approved):

PR #1136, opened 2024-07-01
use generics in SearchApi

### LINKED ISSUE #1134: Use generics in globalsearch
globalSearch currently returns a wildcard which always requires a cast afterwards, even though the return type is [known for each SearchScope](https://github.com/gitlab4j/gitlab4j-api/blob/e8a261a57b87ef5a9eeb44247177443c12f6c09e/src/main/java/org/gitlab4j/api/SearchApi.java#L71).

```
List<?> projects = gitLabApi.getSearchApi().globalSearch(SearchScope.PROJECTS, "text-to-search-for");
```

Ideally we would have a generic return type, based on SearchScope:

```
Pager<SearchScope> search = gitlabApi.searchApi.globalSearch(Constants.SearchScope.BLOBS, "hello")
```

[SearchScope](https://github.com/gitlab4j/gitlab4j-api/blob/e8a261a57b87ef5a9eeb44247177443c12f6c09e/src/main/java/org/gitlab4j/api/Constants.java#L733) is currently defined  as an enum which makes it impossible to use generics. We could convert it to class with static members to be able to use the same syntax and profit from generics.

```java
   // with this definition the above usage of globalSearch compiles:

    public <T> Pager<T> globalSearch(SearchScope<T> searchConstant) {
        // ...
    }

    public static class SearchScope<T> {
        private SearchScope(Class<T> clazz, String jsonName) {
        }
        
        public static final SearchScope<Project> PROJECTS = new SearchScope<>(Project.class, "projects");
        public static final SearchScope<SearchBlob> BLOBS = new SearchScope<>(SearchBlob.class, "blobs");
        
       // JsonCreator, JsonValue and ToString can be implemented using the `jsonName` field
    }

```

This would be a breaking change because code will have to be recompiled, although there should be no changes necessary on the users side.

Of course we would have to implement

exaV:
as proposed in https://github.com/gitlab4j/gitlab4j-api/issues/1134

Fixes #1134

### DIFF
--- a/README.md
+++ b/README.md
@@ -636,7 +636,7 @@ List<Runner> runners = gitLabApi.getRunnersApi().getAllRunners();
 #### SearchApi
 ```java
 // Do a global search for Projects
-List<?> projects = gitLabApi.getSearchApi().globalSearch(SearchScope.PROJECTS, "text-to-search-for");
+List<Project> projects = gitLabApi.getSearchApi().globalSearch(SearchScope.PROJECTS, "text-to-search-for");
 ```
 
 #### ServicesApi
--- a/src/main/java/org/gitlab4j/api/Constants.java
+++ b/src/main/java/org/gitlab4j/api/Constants.java
@@ -1,8 +1,20 @@
 package org.gitlab4j.api;
 
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
+import org.gitlab4j.api.models.Commit;
+import org.gitlab4j.api.models.Issue;
+import org.gitlab4j.api.models.MergeRequest;
+import org.gitlab4j.api.models.Milestone;
+import org.gitlab4j.api.models.Note;
+import org.gitlab4j.api.models.Project;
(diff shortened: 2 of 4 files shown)


exaV:
It should work but I was unable to get the integration test to work. It fails with (even without my changes)
```
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.391 s <<< FAILURE! - in org.gitlab4j.api.TestSearchApi
org.gitlab4j.api.TestSearchApi  Time elapsed: 0.391 s  <<< ERROR!
org.gitlab4j.api.GitLabApiException: invalid_grant

```

exaV:
Okay I was able to get the tests to work. After my latest change all tests in `TestSearchApi` are green. 

To get it to run I had to remove `seedData()` method in `org.gitlab4j.api.SetupIntegrationTestExtension#beforeAll` in my local copy of the code.

exaV:
@jmini please review

exaV:
@jmini Please let me know if you need me to change something to get this merged

jmini:
Yes it is still on my TODO list to review this. Sorry for the delay.

jmini:

Issue #1155, opened 2024-08-25
5.6.0 includes a breaking change in the API


MarkEWaite:
I believe that the API change of the fourth argument `boolean` (primitive type, nulls not allowed) to `Boolean` (object, nulls allowed) is a breaking change in pull request:

* https://github.com/gitlab4j/gitlab4j-api/pull/1123

The change in [the diff](https://github.com/gitlab4j/gitlab4j-api/pull/1123/files#diff-9e127234e0f2bd8027b49f6e23304c0dd323640d2a9847a027ebae9d882fc7c7R2214) went from this:

```
    public ProjectHook addHook(Object projectIdOrPath, String url, ProjectHook enabledHooks,
            boolean enableSslVerification, String secretToken) throws GitLabApiException {
```

to this:

```
    public ProjectHook addHook(Object projectIdOrPath, String url, ProjectHook enabledHooks,
            Boolean enableSslVerification, String secretToken) throws GitLabApiException {
```

The Jenkins issue tracker received two reports of method not found exceptions in issue [JENKINS-73672](https://issues.jenkins.io/browse/JENKINS-73672).  

Is there a critical need for the change of type in the argument?

If not, could that change be reverted and a new release provided so that API compatibility is retained?

basil:
Note that as of https://github.com/jenkinsci/gitlab-branch-source-plugin/pull/438 consumers are actually depending on the _new_ signature 😬

MarkEWaite:
Closing this issue because the Jenkins plugin has updated to the 5.6.0 API and a revert to the 5.5.0 compatible API would be another breaking change.

PR #1156, opened 2024-08-27
Restore API compatibility with 5.5.0 and earlier releases

MarkEWaite:
## Restore API compatibility with 5.5.0 and earlier releases

https://github.com/gitlab4j/gitlab4j-api/commit/224f4c5b9f15d54ad3efa6702009e6d99c0445fd#commitcomment-145297850 notes that the change from boolean to Boolean is a breaking change in the API.

It seems easiest and safest to restore API compatibility rather than requiring lock-step upgrades from consumers of the API.  Jenkins (as an example) allows users to upgrade the GitLab API plugin without requiring that they upgrade the GitLab branch source plugin.  When that happens, the breaking change in the API causes runtime failures for Jenkins users.

https://github.com/jenkinsci/gitlab-branch-source-plugin/issues/435 reports some of those runtime failures.

[JENKINS-73672](https://issues.jenkins.io/browse/JENKINS-73672) reports more runtime failures due to the breaking change in the API.

Fixes https://github.com/gitlab4j/gitlab4j-api/issues/1155

### DIFF
--- a/src/main/java/org/gitlab4j/api/ProjectApi.java
+++ b/src/main/java/org/gitlab4j/api/ProjectApi.java
@@ -2213,7 +2213,7 @@ public class ProjectApi extends AbstractApi implements Constants {
      * @throws GitLabApiException if any exception occurs
      */
     public ProjectHook addHook(Object projectIdOrPath, String url, ProjectHook enabledHooks,
-            Boolean enableSslVerification, String secretToken) throws GitLabApiException {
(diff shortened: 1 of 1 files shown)


basil:
@MarkEWaite Note that as of https://github.com/jenkinsci/gitlab-branch-source-plugin/pull/438 we're actually depending on the _new_ signature :grimacing:

MarkEWaite:
Closing this pull request so that it won't be merged.

PR #1181, opened 2024-10-13
feat: create group webhook #1173

### LINKED ISSUE #1173: No support for Group webhooks
I have not found support for https://docs.gitlab.com/ee/api/group_webhooks.html#add-a-group-hook

Does it exist or is the implementation incomplete?

LeJeanbono:
Fix #1173


src/main/java/org/gitlab4j/api/models/Webhook.java
```diff
@@ -0,0 +1,252 @@
+package org.gitlab4j.api.models;
+
+import java.io.Serializable;
+import java.util.Date;
+
+public class Webhook implements Serializable {
```
jmini, line 6: Probably the class should be named `GroupHook` to be consistent with `ProjectHook` (and both classes could have a common ancestor, but this can be done in a latter step)

LeJeanbono: Done :)

PR #1187, opened 2024-10-28
Add missing deployment and enviorment attributes

katrinSaleschus:


src/main/java/org/gitlab4j/api/models/Environment.java
jmini, line 1: Please update the corresponding json file (add a realistic example value): [environment.json](https://github.com/gitlab4j/gitlab4j-api/blob/c393f61a5aec8cb916da6b19b974ee67ba800979/src/test/resources/org/gitlab4j/api/environment.json)

This is tested by this test:
https://github.com/gitlab4j/gitlab4j-api/blob/c393f61a5aec8cb916da6b19b974ee67ba800979/src/test/java/org/gitlab4j/api/TestGitLabApiBeans.java#L171-L175

src/main/java/org/gitlab4j/api/webhook/DeploymentEvent.java
jmini, line 1: There is not test, but this should be tested in [`TestGitLabApiEvents`](https://github.com/gitlab4j/gitlab4j-api/blob/c393f61a5aec8cb916da6b19b974ee67ba800979/src/test/java/org/gitlab4j/api/TestGitLabApiEvents.java)

jmini (changes requested):
Some tests should be added

katrinSaleschus:
Thank you for your Review. Now everything should be ok to merge.

jmini:
I will have a look

PR #1188, opened 2024-11-01
Add support for Job Token

### LINKED ISSUE #678: Support for CI/CD job tokens?
I was wondering how to use this library with [GitLab CI/CD job tokens](https://docs.gitlab.com/ee/api/#gitlab-cicd-job-token). Does currently not seem to be supported, correct?

I don't find any reference anymore in the GitLab documentation but to authenticate against the API with job tokens you need to use the header `JOB-TOKEN` instead of `PRIVATE-TOKEN`.

Side note: there's [`TokenType.ACCESS`](https://github.com/gitlab4j/gitlab4j-api/blob/master/src/main/java/org/gitlab4j/api/Constants.java#L39) but I did not find any reference to it (unused?).

jmini:
Fixes #678

See: https://docs.gitlab.com/ee/ci/jobs/ci_job_token.html#to-authenticate-a-rest-api-request

### DIFF
--- a/src/main/java/org/gitlab4j/api/Constants.java
+++ b/src/main/java/org/gitlab4j/api/Constants.java
@@ -38,6 +38,7 @@ public interface Constants {
     public enum TokenType {
         ACCESS,
         OAUTH2_ACCESS,
+        JOB_TOKEN,
         PRIVATE;
     }
 
--- a/src/main/java/org/gitlab4j/api/GitLabApiClient.java
+++ b/src/main/java/org/gitlab4j/api/GitLabApiClient.java
@@ -56,6 +56,7 @@ import org.glassfish.jersey.media.multipart.file.StreamDataBodyPart;
 public class GitLabApiClient implements AutoCloseable {
 
     protected static final String PRIVATE_TOKEN_HEADER = "PRIVATE-TOKEN";
+    protected static final String JOB_TOKEN_HEADER = "JOB-TOKEN";
     protected static final String SUDO_HEADER = "Sudo";
     protected static final String AUTHORIZATION_HEADER = "Authorization";
     protected static final String X_GITLAB_TOKEN_HEADER = "X-Gitlab-Token";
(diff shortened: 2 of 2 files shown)


jmini:
Tested with [jbang](https://www.jbang.dev/):

```java
///usr/bin/env jbang "$0" "$@" ; exit $?

//DEPS https://github.com/jmini/gitlab4j-api/commit/17020801279f8e1214dd24ccb0dc6a0401575d61

import java.util.List;
import java.util.logging.Level;

import org.gitlab4j.api.Constants.TokenType;
import org.gitlab4j.api.GitLabApi;
import org.gitlab4j.api.GitLabApi.ApiVersion;
import org.gitlab4j.api.models.Pipeline;
import org.gitlab4j.api.models.Release;

public class HelloWorld {

    public static void main(String... args) throws Exception {
        System.out.println("--- Start ---");
        String gitLabUrl = getEnv("CI_API_V4_URL");
        if (gitLabUrl.endsWith("/api/v4")) {
            gitLabUrl = gitLabUrl.substring(0, gitLabUrl.length() - "/api/v4".length());
        }
        String jobToken = getEnv("CI_JOB_TOKEN");
        String projectId = getEnv("CI_PROJECT_ID");
        try (GitLabApi gitLabApi = new GitLabApi(ApiVersion.V4, gitLabUrl, TokenType.JOB_TOKEN, jobToken)
                .withRequestResponseLogging(Level.INFO)) {
            List<Release> releases = gitLabApi.getReleasesApi()
                    .getReleases(projectId);
            System.out.println(releases);

            Pipeline pipeline = gitLabApi.getPipelineApi()
                    .triggerPipeline(projectId, jobToken, "main", null);
            System.out.println(pipeline);
        }
        System.out.println("--- End ---");
    }

    private static String getEnv(String key) {
        String gitLabUrl = System.getenv(key);
        if (gitLabUrl == null) {
            System.out.println(key + " is not defined, please define it");
            System.exit(1);
            throw new RuntimeException("unreachable code");
        }
        return gitLabUrl;
    }
}
```

PR #1198, opened 2024-11-24
Object Attributes

bdgould:
I keep getting MergeRequestEvent having a null objectAttributes field.  It looks like the API actually uses the syntax: object_attributes.  This seems to fix it.  There could be other instances of elsewhere in the API.

I'm trying create a simple webhook trigger using a managed GitLab CE 17.5.1 server, and am getting event json like:  [sample.json](https://github.com/user-attachments/files/17894370/sample.json)

### DIFF
--- a/src/main/java/org/gitlab4j/api/webhook/ExternalStatusCheckEvent.java
+++ b/src/main/java/org/gitlab4j/api/webhook/ExternalStatusCheckEvent.java
@@ -7,6 +7,8 @@ import org.gitlab4j.api.models.Assignee;
 import org.gitlab4j.api.utils.JacksonJson;
 import org.gitlab4j.api.webhook.MergeRequestEvent.ObjectAttributes;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 public class ExternalStatusCheckEvent implements Serializable {
     private static final long serialVersionUID = 1L;
 
@@ -15,7 +17,10 @@ public class ExternalStatusCheckEvent implements Serializable {
     private EventUser user;
     private EventProject project;
     private EventRepository repository;
+
+    @JsonProperty(value = "object_attributes")
     private ObjectAttributes objectAttributes;
+
     private List<EventLabel> labels;
     private MergeRequestChanges changes;
     private List<Assignee> assignees;
--- a/src/main/java/org/gitlab4j/api/webhook/IssueEvent.java
+++ b/src/main/java/org/gitlab4j/api/webhook/IssueEvent.java
@@ -5,6 +5,8 @@ import java.util.List;
 import org.gitlab4j.api.models.Assignee;
 import org.gitlab4j.api.utils.JacksonJson;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 public class IssueEvent extends AbstractEvent {
     private static final long serialVersionUID = 1L;
 
@@ -14,7 +16,10 @@ public class IssueEvent extends AbstractEvent {
     private EventUser user;
     private EventProject project;
     private EventRepository repository;
+
+    @JsonProperty(value = "object_attributes")
     private ObjectAttributes objectAttributes;
+
     private List<Assignee> assignees;
     private Assignee assignee;
     private List<EventLabel> labels;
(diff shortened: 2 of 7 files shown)


jmini:
How are you using this library?

Because the convention of turning underscore notation to camelCase should work out of the box, without us having to redefine it everywhere using a `@JsonProperty` property annotation.

This is also covered by Unit Tests here:
https://github.com/gitlab4j/gitlab4j-api/blob/0a1b6e0dcc926bb9fa1fb284d12b89388f446548/src/test/java/org/gitlab4j/api/TestGitLabApiEvents.java

So I guess that you have a different jackson config (maybe a different ObjectMapper instance).

jmini:
When I have deleted the `6.x` branch (see https://github.com/gitlab4j/gitlab4j-api/issues/926#issuecomment-2506771237) this PR was closed, but this is not my intention. It can be changed to target `main` (which is the branch where the next `6.0.0` version is prepared).

jmini:
When I have deleted the `6.x` branch (see https://github.com/gitlab4j/gitlab4j-api/issues/926#issuecomment-2506771237) this PR was closed, but this is not my intention. It can be changed to target `main` (which is the branch where the next `6.0.0` version is prepared).

jmini:
I think the configuration is done here:

https://github.com/gitlab4j/gitlab4j-api/blob/e71f22bae191d10ba8b0fffe1df1a571f7285917/gitlab4j-api/src/main/java/org/gitlab4j/api/utils/JacksonJson.java#L63

To configure the ObjectMapper that does the case transformation inside a Quarkus project you need to do this:

```java
import javax.enterprise.inject.Instance;
import javax.inject.Singleton;

import org.gitlab4j.api.utils.JacksonJson;

import com.fasterxml.jackson.databind.ObjectMapper;
import com.fasterxml.jackson.databind.SerializationFeature;

import io.quarkus.jackson.ObjectMapperCustomizer;

public class ObjectMapperConfiguration {

	@Singleton
	ObjectMapper objectMapper(Instance<ObjectMapperCustomizer> customizers) {
		return new JacksonJson().getObjectMapper().enable(SerializationFeature.INDENT_OUTPUT);
	}
}
```

I am almost certain other frameworks have similar approaches.

jmini (changes requested):
I am not against having more information in Annotations on the class, to make it independent from a potential Jackson config.

To accept this Pull Request, work is needed:

1. add the annotation everywhere needed (not just on few fields) --> consistency is very important
2. modify the test so that they run without this specific Jackson config (and they have to stay green).

bdgould:
Hi @jmini , Sorry for the delay!  I agree, I think the metadata would be helpful -- I was just tied up during the Thanksgiving holiday here in the U.S. and then some unrelated stuff.  I can work through it a bit more here in the coming weeks!

jmini:
What a work! I will review ASAP and merge.

bdgould:
Ha, thanks @jmini.  I may need another weekend or two to test a bit more.  I tried a couple different things with our managed gitlab server and found a couple date formats were wrong -- I'd like to try a test a bunch more!

jmini:
@bdgould it could even be that some date in the test JSON files in the [`gitlab4j-models/src/test/resources/org/gitlab4j/models`](https://github.com/gitlab4j/gitlab4j-api/tree/7f6aa16d918d068a97c1de5c9aa0d2559843b92d/gitlab4j-models/src/test/resources/org/gitlab4j/models) folder not really matches what gitlab is sending...

bdgould:
Yeah, I found a couple of slight issues like that.  I'm just trying various things with my managed gitlab instance to verify so far.

jmini:
I fixed some more cases where the date is sent as `2018-07-01` by the server and where we were serializing as `2018-07-01T00:00:00Z` in PR https://github.com/gitlab4j/gitlab4j-api/pull/1223

jmini:
@katrinSaleschus I think the work was never finished.

As discussed, I am not opposed to use Jackson annotation everywhere instead of specific ObjectMapper configuration. But the work needs to be complete and free of failure in the unit test.

Since this work was started there were a lot of changes, you need to restart where @bdgould left, fix the merge conflict and finish the open tasks.

PR #1215, opened 2024-12-20
Add work item support to WebHookManager

ParkerMc:
It appears that the WorkItemEvent has been implemented but not added to the WebHookManager. With gitlab 17.7 new issues and epics use an object_kind of work_item when webhooks are sent.


jmini (approved):
I am not really using the `WebHookManager` myself so I missed it when implementing the `WorkItemEvent`.

 I think this addition makes sense, thank you for this pull-request.

jmini:
🎉 merged!

Maybe `WorkItemEvent` also needs to be updated after the GitLab update 17.7 (I didn't tested yet with this version)

PR #1233, opened 2025-03-05
Add JobApi#getJob()

### LINKED ISSUE #1231: Support https://docs.gitlab.com/api/jobs/#get-job-tokens-job in JobApi
The actual gitlab v4 API allows to fetch jobs by token https://docs.gitlab.com/api/jobs/#get-job-tokens-job
```
curl --header "Authorization: Bearer $CI_JOB_TOKEN" "${CI_API_V4_URL}/job"
curl --header "JOB-TOKEN: $CI_JOB_TOKEN" "${CI_API_V4_URL}/job"
curl "${CI_API_V4_URL}/job?job_token=$CI_JOB_TOKEN"
```

But the Java API does not have any way to access that endpoint.

Implementation would be roughly:
```
/**
     * GET /job
     * Examples (must run as part of the script section of a CI/CD job):
     * curl "${CI_API_V4_URL}/job?job_token=$CI_JOB_TOKEN"
     */
    public Job getJob(final String ciJobToken) throws GitLabApiException {
        // this method is currently not exposed in JobApi
        final Response response = get(
                Response.Status.OK, 
                new GitLabApiForm().withParam("job_token", ciJobToken).asMap(), 
                "job");
        return (response.readEntity(new GenericType<Job>() {}));
    }
```

However, we had trouble implementing it like this, because the authorization via the job_token and the authorization via the personal access token seemed to conflict with each other.

jmini:
See documentation: https://docs.gitlab.com/api/jobs/#get-job-tokens-job

Fixes #1231

### DIFF
--- a/gitlab4j-api/src/main/java/org/gitlab4j/api/JobApi.java
+++ b/gitlab4j-api/src/main/java/org/gitlab4j/api/JobApi.java
@@ -292,6 +292,20 @@ public class JobApi extends AbstractApi implements Constants {
     }
 
     /**
+     * Retrieve the job corresponding to the <code>$CI_JOB_TOKEN</code> environment variable (Using a {@link org.gitlab4j.models.Constants.TokenType#JOB_TOKEN} authentication).
+     *
+     * <pre><code>GitLab Endpoint: GET /job</code></pre>
+     *
+     * @return a single job
+     * @throws GitLabApiException if any exception occurs during execution
+     */
+    public Job getJob() throws GitLabApiException {
+        Response response =
+                get(Response.Status.OK, null, "job");
+        return (response.readEntity(Job.class));
+    }
+    
+    /**
      * Get single job in a project.
      *
      * <pre><code>GitLab Endpoint: GET /projects/:id/jobs/:job_id</code></pre>
--- a/gitlab4j-api/src/main/java/org/gitlab4j/api/JobApi.java
+++ b/gitlab4j-api/src/main/java/org/gitlab4j/api/JobApi.java
@@ -300,11 +300,10 @@ public class JobApi extends AbstractApi implements Constants {
      * @throws GitLabApiException if any exception occurs during execution
      */
     public Job getJob() throws GitLabApiException {
-        Response response =
-                get(Response.Status.OK, null, "job");
+        Response response = get(Response.Status.OK, null, "job");
         return (response.readEntity(Job.class));
     }
-    
+
     /**
      * Get single job in a project.
      *


gitlab4j-api/src/main/java/org/gitlab4j/api/JobApi.java
```diff
@@ -291,6 +291,19 @@ public Stream<Job> getJobsStream(Object projectIdOrPath, long pipelineId, Boolea
...
+    /**
+     * Retrieve the job corresponding to the <code>$CI_JOB_TOKEN</code> environment variable (Using a {@link org.gitlab4j.models.Constants.TokenType#JOB_TOKEN} authentication).
+     *
+     * <pre><code>GitLab Endpoint: GET /job</code></pre>
+     *
+     * @return a single job
+     * @throws GitLabApiException if any exception occurs during execution
+     */
+    public Job getJob() throws GitLabApiException {
+        Response response = get(Response.Status.OK, null, "job");
```
tkruse, line 303: I would recommend checking that a JOB_TOKEN was set in gitlabApi and throwing a meaningful exception if not.

gitlab4j-api/src/main/java/org/gitlab4j/api/JobApi.java
```diff
@@ -291,6 +291,19 @@ public Stream<Job> getJobsStream(Object projectIdOrPath, long pipelineId, Boolea
         return (getJobsForPipeline(projectIdOrPath, pipelineId, getDefaultPerPage(), includeRetried).stream());
     }
 
+    /**
+     * Retrieve the job corresponding to the <code>$CI_JOB_TOKEN</code> environment variable (Using a {@link org.gitlab4j.models.Constants.TokenType#JOB_TOKEN} authentication).
+     *
+     * <pre><code>GitLab Endpoint: GET /job</code></pre>
+     *
+     * @return a single job
```
tkruse, line 299: I would recommend adding more explanation that this will only work while the CI_JOB_TOKEN is valid (while the CI job is running) or similar.

gitlab4j-api/src/main/java/org/gitlab4j/api/JobApi.java
jmini: Added with https://github.com/gitlab4j/gitlab4j-api/pull/1236

gitlab4j-api/src/main/java/org/gitlab4j/api/JobApi.java
jmini: See https://github.com/gitlab4j/gitlab4j-api/pull/1236

PR #1236, opened 2025-03-10
Add JobApi#getJob(String)

jmini:
Follow up of https://github.com/gitlab4j/gitlab4j-api/pull/1233 to add the the `getJob()` method where the ci-token can be passed as query parameter.

### DIFF
--- a/gitlab4j-api/src/main/java/org/gitlab4j/api/JobApi.java
+++ b/gitlab4j-api/src/main/java/org/gitlab4j/api/JobApi.java
@@ -292,15 +292,35 @@ public class JobApi extends AbstractApi implements Constants {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


gitlab4j-api/src/main/java/org/gitlab4j/api/JobApi.java
```diff
@@ -292,18 +292,38 @@ public Stream<Job> getJobsStream(Object projectIdOrPath, long pipelineId, Boolea
...
+        if (tokenType != TokenType.JOB_TOKEN) {
+            throw new IllegalStateException(
+                    "This method can only be called with a " + TokenType.JOB_TOKEN + " authentication");
+        }
         Response response = get(Response.Status.OK, null, "job");
         return (response.readEntity(Job.class));
     }
 
+    /**
+     * Retrieve the job corresponding to the <code>$CI_JOB_TOKEN</code> environment variable.
```
jmini, line 313: This works only without any authentication, maybe this information would be needed in the javadoc

PR #1237, opened 2025-03-28
Added missing attributes for ProjectApi

yarisvt:
Added missing attributes to the ProjectApi


gitlab4j-models/src/main/java/org/gitlab4j/api/models/Project.java
```diff
@@ -789,7 +828,7 @@ public void setBuildCoverageRegex(String buildCoverageRegex) {
 
     public Project withBuildCoverageRegex(String buildCoverageRegex) {
         this.buildCoverageRegex = buildCoverageRegex;
-        return this;
+        return (this);
```
jmini, line 831: I don't see why we need those brackets.

I would just do this:

```suggestion
        return this;
```

gitlab4j-models/src/main/java/org/gitlab4j/api/models/Project.java
jmini, line 1: Can you update the file `project.json`?
https://github.com/gitlab4j/gitlab4j-api/blob/fe14e34f0ab11e9f06f52765f86fbc0d72caa3dd/gitlab4j-models/src/test/resources/org/gitlab4j/models/project.json
Ideally with values coming from an existing from a real REST call.

This way we would we would ensure that your changes are correctly serialized and deserialized.

jmini (changes requested):
This looks like a very good addition.

The example file used in unit-test should be updated.

gitlab4j-models/src/main/java/org/gitlab4j/api/models/Project.java
yarisvt: Some properties are only configurable with a Premium/Ultimate gitlab server, which I unfortenately do not have access to, so I've used an accepted value based on the documentation:
|option                                      |type   |value I used|
|--------------------------------------------|-------|------------|
|use_custom_template                         |boolean|false       |
|external_authorization_classification_label |string |"label"       |
|mirror_trigger_builds                       |boolean|false       |
|only_allow_merge_if_all_status_checks_passed|boolean|false       |
|group_with_project_templates_id             |integer|1           |

gitlab4j-models/src/test/resources/org/gitlab4j/models/project.json
```diff
@@ -126,5 +126,40 @@
...
+  "group_runners_enabled" : true,
+  "show_default_award_emojis" : false,
+  "warn_about_potentially_unwanted_characters" : true,
+  "mirror_trigger_builds" : false,
+  "auto_cancel_pending_pipelines" : "enabled",
+  "repository_object_format" : "sha1",
+  "only_allow_merge_if_all_status_checks_passed" : false,
+  "avatar" : null,
+  "group_with_project_templates_id" : null,
+  "public_builds" : null,
```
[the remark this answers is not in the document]
yarisvt: Right, I missed those. I have also removed the `avatar` property, since that can only be used with a multipart/form-data and I now saw that there already is a `setProjectAvatar` method that uses this property

jmini (approved):

gitlab4j-models/src/main/java/org/gitlab4j/api/models/ContainerExpirationPolicy.java
```diff
@@ -4,15 +4,26 @@
...
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 public class ContainerExpirationPolicy implements Serializable {
     private static final long serialVersionUID = 1L;
 
     private String cadence;
     private Boolean enabled;
+
+    @JsonProperty("keep_n")
```
jmini, line 15: Currently the way the models are written, we are not using `@JsonProperty`. There was an attempt in https://github.com/gitlab4j/gitlab4j-api/pull/1198, but it never was finished.

For consistency reason, I would not add this now.

gitlab4j-api/src/main/java/org/gitlab4j/api/ProjectApi.java
```diff
@@ -4727,4 +4728,25 @@ public List<Iteration> listProjectIterations(Object projectIdOrPath, IterationFi
...
+     *
+     * <pre><code>GitLab Endpoint: PUT /projects/:id</code></pre>
+     *
+     * @param projectIdOrPath the project in the form of an Long(ID), String(path), or Project instance, required
+     * @param containerExpirationPolicyAttributes the ContainerExpirationPolicyAttributes instance
+     * @return the updated Project instance
+     * @throws GitLabApiException if any exception occurs
+     */
+    public Project setProjectContainerExpirationPolicy(
+            Object projectIdOrPath, ContainerExpirationPolicyAttributes containerExpirationPolicyAttributes)
```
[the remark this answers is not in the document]
yarisvt: I tried that before, but it does not work, I keep getting this error: `container_expiration_policy_attributes is invalid`

According to the documentation, `container_expiration_policy_attributes` is used in the [create](https://docs.gitlab.com/api/projects/#create-a-project)/[edit](https://docs.gitlab.com/api/projects/#edit-a-project) API to modify the policy, while `container_expiration_policy` is used in the list API to [list](https://docs.gitlab.com/api/projects/#get-a-single-project) the policy.

Also, there is an [example](https://docs.gitlab.com/user/packages/container_registry/reduce_container_registry_storage/#use-the-cleanup-policy-api) which updates the policy, which uses `--header 'Content-Type: application/json;charset=UTF-8'`

I think it doesn't work with `withContainerExpirationPolicy(..)`. since that request eventually uses  `--header Content-Type: application-x-www-form-urlencoded`. But the `container_expiration_policy_attributes` is formatted as JSON: 
```json
{
  "cadence" : "1month",
  "enabled" : true,
  "keep_n" : 100,
  "older_than" : "7d",
  "name_regex" : ".*-development",
  "name_regex_keep" : ".*-main"
}
```

jmini:

Issue #1104, opened 2024-03-19
JobScope missing value


ramarro123:
hi, i think that JobScope missing scheduled value, for scheduled job that run with a delay

am i wrong?

jmini:
Probably, often the models in this project are a little behind the official API. Can you point to the GitLab documentation?

A PR to add the missing field is welcome.

fasilmarshooq:
I have similar reuest for scope `waiting_for_resource` , MR -> https://github.com/gitlab4j/gitlab4j-api/pull/1239

PR #1239, opened 2025-04-07
added  WAITING_FOR_RESOURCE to job scopes

fasilmarshooq:
I need this scope to track the jobs queued in the resource group.

MR also relates to https://github.com/gitlab4j/gitlab4j-api/issues/1104


jmini:
From the docs https://docs.gitlab.com/api/jobs/

> Scope of jobs to show. Either one of or an array of the following: `created`, `pending`, `running`, `failed`, `success`, `canceled`, `skipped`, `waiting_for_resource`, or `manual`

So I think this is a good addition 👍

PR #1240, opened 2025-04-10
Added missing attributes for webhook events

yarisvt:
Updated the following events with missing properties
- `AbstractPushEvent`
- `BuildCommit`
- `EventCommit`
- `EventIssue`
- `JobEvent`
- `PipelineEvent`
- `WikiPageEvent`

Especially, `PipelineEvent` and `JobEvent` have changed.

`PipelineEvent`  no longer uses a `jobs` array (`Job`), but a `builds` array (`Build`). I have not removed the `jobs` array for backwards compatibility.

`JobEvent` no longer uses `job` prefixed properties, but `build` prefixed properties. I have not removed the `job` prefixed properties for backwards compatibility. I did change the object_kind to `build` as it should be according to the docs (used to be `job` in version 12.x, but changed to `build` in version 13.x). This might break backwards compatibility.

### DIFF
--- a/gitlab4j-models/src/main/java/org/gitlab4j/api/models/BuildStatus.java
+++ b/gitlab4j-models/src/main/java/org/gitlab4j/api/models/BuildStatus.java
@@ -1,0 +1,60 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Greg Messner <greg@messners.com>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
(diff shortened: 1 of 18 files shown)


gitlab4j-models/src/main/java/org/gitlab4j/api/models/BuildStatus.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.gitlab4j.models.utils.JacksonJsonEnumHelper;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * Enum for the various Build status values.
+ */
+public enum BuildStatus {
```
jmini, line 34: I am wondering if this is not a duplication of the enum `Constants.JobScope` that we already have.

But `BuildStatus` seems a better name and feels more natural.

jmini (approved):
This looks a very good addition to the project. 👍

jmini (changes requested):
Just there are build failure after your change.

I think this is due to the fact that there was already a `BuildEvent` with OBJECT_KIND `build`

https://github.com/gitlab4j/gitlab4j-api/blob/7e8795e75b83312cbba9bd70e396587969468758/gitlab4j-api/src/main/java/org/gitlab4j/api/WebHookManager.java#L179-L190

I didn't look into all the details, but different/more cleanup might be necessary.

yarisvt:
You're right, there already is a `BuildEvent`,  I missed that somehow. I've reverted the changes in `JobEvent`, and updated `BuildEvent` with some missing properties.

gitlab4j-models/src/main/java/org/gitlab4j/api/models/BuildStatus.java
yarisvt: I actually copied `JobStatus` into `BuildStatus`, they're the same except for the name. But I figured it's more clear to distinguish between jobs and builds

jmini (approved):
No problem.

Thank you very much for the update and the 🧹 .

PR #1243, opened 2025-04-17
Add associations count for users

sichapman:
Based on https://docs.gitlab.com/api/users/#get-a-count-of-a-users-projects-groups-issues-and-merge-requests


jmini (changes requested):
This looks like a good addition.

Can you please put an example JSON file example (ideally coming from a running GitLab installation, or coming from the GitLab documentation) in `gitlab4j-models/src/test/resources/org/gitlab4j/models`

And add a new test case like this:
https://github.com/gitlab4j/gitlab4j-api/blob/7e8795e75b83312cbba9bd70e396587969468758/gitlab4j-models/src/test/java/org/gitlab4j/models/TestGitLabApiBeans.java#L44-L48

To make sure that the new `Associations` model can be serialised and deserialised as expected?

jmini (approved):

jmini:
🎉 I have added the missing test and test file, and I have merged your PR. Thank you for this contribution!

PR #1269, opened 2025-08-11
Expose ci_delete_pipelines_in_seconds via Project API

lheer:
Since gitlab 17.9, it is possible to configure a project variable after which project pipelines get deleted automatically.

This commit exposes this setting via the Project API.


jmini (approved):
Looks good to me

jmini:
🎉 thank you very much for this contribution!

PR #1277, opened 2025-09-02
Bugfix/merge request event object attributes

katrinSaleschus:
This is my contribution to finish the great work of @bdgould in Object Attributes #1198


jmini:
I need to figure out your change and understand it.

I am surprised by the change of the tests and the dates. I understand the GitLab is not consistent but those example json files are supposed to represent "real" JSON responses sent by GitLab. If you modify those, how are we making sure that this client continues to parse those JSON responses correctly?

katrinSaleschus:
I changed the test data because it didn't match what I received from GitLab. And you mentioned [here](https://github.com/gitlab4j/gitlab4j-api/pull/1198#issuecomment-2560920246) that this could be the case. Since the data from GitLab is inconsistent anyway, I didn't undo the change. But if you want me to, I'm happy to do so. Perhaps it would also make sense to only partially undo the change, as this would best reflect reality.

katrinSaleschus:
I would say the main change to what @bdgould did are the [custom Deserializer](https://github.com/gitlab4j/gitlab4j-api/pull/1277/files#diff-46679c4f9249eed0347c8b3b7666cbb622647731eb91352a1e923f1320e47ad1) to make sure the library can handle multiple date formats.

PR #1288, opened 2025-09-22
feat:added new AccessLevel Planner

ysemko:
added new Role "Planner" introduced in GitLab 17.7:
https://about.gitlab.com/releases/2024/12/19/gitlab-17-7-released/#new-planner-user-role


ysemko:
Closes #1288

PR #1297, opened 2025-10-20
Support the "environment" field in the Pipeline and Job webhook events

lucasls:
The Pipeline and Job webhook events support the `environment` for deployment jobs, but they are missing in the library.

You can find more information at https://docs.gitlab.com/user/project/integrations/webhook_events/ (search for `"environment": {` to see an example).

I noticed the `Build` class is not in the `models` package, instead of `webhook`. I didn't change it since it'd break compatibility, but let me know if you think I should change something about it.

### DIFF
--- a/gitlab4j-models/src/test/resources/org/gitlab4j/models/build-event.json
+++ b/gitlab4j-models/src/test/resources/org/gitlab4j/models/build-event.json
@@ -54,6 +54,11 @@
 		"path_with_namespace": "gitlab-org/gitlab-test",
 		"default_branch": "master"
 	},
+    "environment": {
+        "name": "Test Env",
+        "action": "start",
+        "deployment_tier": "testing"
+    },
 	"runner": {
 		"active": true,
 		"runner_type": "project_type",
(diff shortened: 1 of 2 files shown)


jmini (changes requested):
Looks good to me.

I am not sure why we have 2 packages:

* `org.gitlab4j.api.webhook`
* `org.gitlab4j.api.models`

But since we have them and `Build` is only used by `PipelineEvent`, I agree it should be in `webhook` and not `models`. I agree with your point of not changing it to not break anything.

The example json files that correspond to `PipelineEvent` and `BuildEvent`:

* `"pipeline-event.json"`
* `"build-event.json"`

here: https://github.com/gitlab4j/gitlab4j-api/tree/main/gitlab4j-models/src/test/resources/org/gitlab4j/models

--> They should be updated (they should contain the example values corresponding to the model extension made with this PR)

gitlab4j-models/src/main/java/org/gitlab4j/api/webhook/BuildEvent.java
```diff
@@ -42,6 +43,8 @@ public class BuildEvent extends AbstractEvent {
     private EventProject project;
     private Runner runner;
 
+    private Environment environment;
```
jmini: Shouldn't this be `EventEnvironment` instead of `Environment`?

At least this would match my observations…

jmini (approved):
LGTM. Thank you!

PR #1194, opened 2024-11-18
Add missing attributes for NotesApi and DiscussionApi (#1193)

MadnessIRL:
Add attributes "created_at", "internal" for NotesApi commits and DiscussionApi merge requests

- Replace unmappable characters in javadocs

Fixes: #1193

PR #1305, opened 2025-12-04
[5.x.x] NotesApi: fix createMergeRequestNote() API breakage introduced in 5.8.0

peat-psuwit:
Even when the new `created_at` and `internal` arguments are marked as "optional", that doesn't mean one can simply omit them from argument list. This means `createMergeRequestNote()`'s function signature has been changed, which is a breaking change which happens without major version bump.

So add a wrapper function with the old signature that calls the new signature, which restores compatibility to un-migrated code.

IMO this change is not needed for 6.x.x line, as that can be seen as major update and API breakage is fine that way. Although, I'll admit that with 5.8.0 released over a year ago, I'm not sure if this still matters anymore...

Fixes: 60259a69da ("Add attributes "created_at", "internal" for NotesApi commits and DiscussionApi merge requests (#1194)")

### DIFF
--- a/src/main/java/org/gitlab4j/api/NotesApi.java
+++ b/src/main/java/org/gitlab4j/api/NotesApi.java
@@ -525,6 +525,22 @@ public class NotesApi extends AbstractApi {
     }
 
     /**
+     * Create a merge request's note.
+     *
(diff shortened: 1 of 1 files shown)


jmini:
I am not really considering maintaining the 5.x.x branch anymore.

I can merge this one, but I am not really planning a release

PR #1308, opened 2025-12-29
Add getters to ProjectFilter

inoorinoor:
ProjectFilter currently does not expose its internal fields via getters, which makes it difficult to verify filter configuration in tests.

In my tests, I need to ensure that ProjectFilter is created with a specific lastActivityAfter value. A similar approach is already possible with MergeRequestFilter, which provides getters for its fields, but ProjectFilter does not.

This PR adds getters to ProjectFilter to improve testability and make its API more consistent with other filter classes in the project.


jmini (approved):
LGTM!

PR #1316, opened 2026-03-05
Update Jackson dependencies to version 2.18.6

michix:
The problem with version 2.18.0 is that there is a security finding (https://github.com/advisories/GHSA-72hv-8253-57qq). I would appreciate it very much if you would consider releasing a patch version.


michix:
May I kindly ask you to release a new version of the project? The security finding fixed with this PR brings alerts in security scanners. Many thanks in advance!

PR #1317, opened 2026-03-19
Update EventMergeRequest webhook to handle non-string merge param values

klew00:
Calls from GitLab's merge request webhook event can have the `merge_params` property contain non-string values, such as `should_remove_source_branch` which is a boolean or `train_ref` which is an object (https://gitlab.com/gitlab-org/gitlab/-/merge_requests/129979). Example:

```
"object_attributes": {
  ...
  "merge_params": {
    "force_remove_source_branch": "1",
    "auto_merge_strategy": "merge_train",
    "should_remove_source_branch": true,
    "sha": "61f4798ee6668860a47f2f84b31bf8b88ab52bb9",
    "train_ref": {
      "commit_sha": "83d69a142f2fb301c3cce0a802ec1c5ee3a091a8",
      "merge_commit_sha": "83d69a142f2fb301c3cce0a802ec1c5ee3a091a8",
      "squash_commit_sha": "61f4798ee6668860a47f2f84b31bf8b88ab52bb9"
    }
  },
  ...
}
```

This causes issues when handling the event since the `merge_params` attribute in `EventMergeRequest` is a `Map<String, String>`, which expects every value to be a string. This change updates `merge_params` so it can handle the `should_remove_source_branch` and `train_ref` non-string values when converting the JSON.

### DIFF
--- a/gitlab4j-models/src/main/java/org/gitlab4j/api/webhook/EventMergeRequest.java
+++ b/gitlab4j-models/src/main/java/org/gitlab4j/api/webhook/EventMergeRequest.java
@@ -2,11 +2,13 @@ package org.gitlab4j.api.webhook;
 
 import java.util.Date;
 import java.util.List;
-import java.util.Map;
 
 import org.gitlab4j.api.models.Assignee;
 import org.gitlab4j.api.models.Duration;
 import org.gitlab4j.models.utils.JacksonJson;
+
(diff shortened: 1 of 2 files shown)


jmini:
This is a good change.
It creates a breaking change for library consumers, but I think this is fine.

PR #1326, opened 2026-06-22
Support name field in ProjectHook

mhagnumdw:
Expose the GitLab project hook `name` field (getter/setter + fluent `withName`), add fluent `withDescription`, and send the `name` param in ProjectApi.addHook / modifyHook.

Also remove a duplicated deployment_events param in addHook.

doc: https://docs.gitlab.com/api/project_webhooks/


mhagnumdw:
I've already tested it in my actual local environment.

